PR #2, opened 2020-03-20
Code clean

davidzengxhsh:
code clean:
remove useless code; move document to doc;
add gitignore to make a clear view
add a common config.mk for lib/test/samples; 
make the code gcc 4.8.5 compile-able remove the hardcoded advance toolchain;
move library code to separete directory.


Makefile
```diff
@@ -1,47 +1,21 @@
...
-SHAREDSONAMELIB = $(PACKAGENAME).so.$(SOVERSION)
-SHAREDLIB = $(PACKAGENAME).so.$(VERSION)
-LIBLINK = $(PACKAGENAME).so
+all: $(subdirs)
 
-INC = ./inc_nx
+.PHONY: $(subdirs)
+$(subdirs):
+	@if [ -d $@ ]; then				\
+		$(MAKE) -C $@ || exit 1;	 	\
```
rzinsly: I don't think it's a good practice to build the samples and the tests with the library, users usually only want to build the library and will have to wait until the rest is built.

davidzengxhsh: If users want to only build the library, they can cd into the library directory and make. Build the lib, samples, test together can avoid incompatible between lib and sample/test.

rzinsly: @davidzengxhsh Yes, but the root directory is the go to place to build many libraries, so by seeing a Makefile there most people will just try make, and our README implies that you should do that.
The tests are interesting for the developers and the C code on samples is usually enough to understand the usage. I still believe that is better to just build the library in this Makefile, but I'll not block the commit.

PR #10, opened 2020-04-15
Test case to recreate kernel oops

rzinsly:
There was a bug in kernel that caused a kernel oops if THP was
enabled.
Set csb_poll_max to 200, test_stress will trigger kernel oops if
the kernel doesn't have fix.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,9 @@ $(subdirs):
 check:  $(subdirs)
 	$(MAKE) -C test $@
 
+unsafe-check:  $(subdirs)
+	$(MAKE) -C test $@
+
 clean:
 	@for dir in $(subdirs); do 			\
 		if [ -d $$dir ]; then			\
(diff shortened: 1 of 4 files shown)


tuliom:
I'm not sure if it's a good idea to cause a kernel oops when running `make check`.
Can we detect this issue and report the failure without causing the kernel oops?
I don't oppose having potentially dangerous tests.  But I think they shouldn't be executed unless the user understands the implication.

rzinsly:
> I'm not sure if it's a good idea to cause a kernel oops when running `make check`.
> Can we detect this issue and report the failure without causing the kernel oops?

I don't know how we can detect this without the failure.

> I don't oppose having potentially dangerous tests. But I think they shouldn't be executed unless the user understands the implication.

It make sense to not execute tests like these with make check, how about a separated script to run this kind of tests?

tuliom:
> It make sense to not execute tests like these with make check, how about a separated script to run this kind of tests?

Yep.  Maybe also an extra target, e.g. `make unsafe-check`

PR #12, opened 2020-04-21
Test if PID is reused while NX window is open

rzinsly:
Adds a new test to check if a PID is being used
while a NX windows is still open with the same PID.

Signed-off-by: Raphael Moreira Zinsly <rzinsly@linux.ibm.com>

### DIFF
--- a/test/test_pid_reuse.c
+++ b/test/test_pid_reuse.c
@@ -1,0 +1,173 @@
+/* Check if there is a bug with the kernel where it reuses a PID of a idle
+ * thread with a VAS window open.
+ * This test opens a VAS window and then fork other process to see if one of
+ * the children can use the same kernel PID of the first one with the window
+ * still open.
+ *
+ * Copyright (C) 2020 IBM Corporation
+ *
+ * Licenses for GPLv2 and Apache v2.0:
+ *
+ * GPLv2:
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ *
+ * Apache v2.0:
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/syscall.h>
+#include <sys/wait.h>
+#include <assert.h>
+#include <errno.h>
+#include <pthread.h>
+#include "zlib.h"
+
+/* Globals shared by all threads. */
+z_stream c_stream;
+
+void* comp_file_multith (void *iterations)
+{
+	int err;
+	long *sleep_sec;
+
+	pid_t my_pid = syscall(SYS_gettid);
+	fprintf(stderr, "thread started: pid: %d\n", (int) my_pid);
+
+	/* Open the window. */
+	err = deflateInit(&c_stream, Z_DEFAULT_COMPRESSION);
+	if (err != 0) {
+		printf("nx_deflateInit err %d\n", err);
+		return (void *) 1;
+	}
+
+	sleep_sec = (long *) iterations;
+	sleep(sleep_sec[0]);
+
+	fprintf(stderr, "thread %d exits\n", (int) my_pid);
+	return (void *) 0;
+}
+
+void fork_children (int parent_pid) {
+	int child_id, i = 0;
+	int seconds = 10;
+	int children = 2048;
+
+	for (i=0; i < children; i++) {
+		child_id = fork();
+
+		if (child_id > 0) {
+			/* PID is being reused. */
+			if (child_id == parent_pid) {
+				fprintf(stderr, "!!! pid %d reused (see ps)\n",
+					child_id);
+				/* Keep this child alive enough time to hit the
+				 * csb error signal. */
+				sleep(300);
+			}
+			continue;
+		} else if (child_id == 0) {
+			pid_t pid = getpid();
+			fprintf(stderr, "%d: child(%d)\n", i, (int) pid);
+			/* Wait enough time to fork all children. */
+			sleep(seconds);
+			fprintf(stderr, "%d: child(%d) exits\n", i, (int) pid);
+			exit(0);
+		} else if (child_id == -1) {
+			fprintf(stderr, "fork failed: %s\n", strerror(errno));
+			exit(-1);
+		}
+	}
+}
+
+int main(int argc, char **argv)
+{
+	int rc, compLen = 1024;
+	char *compr;
+	pthread_t thread;
+	void *ret;
+	long iterations;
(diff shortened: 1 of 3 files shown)


test/test_pid_reuse.c
```diff
@@ -0,0 +1,122 @@
+/* Test if children processes can reuse PID while NX window is open. */
```
tuliom, line 1: Should this file have a licensing text?

test/test_pid_reuse.c
```diff
@@ -0,0 +1,122 @@
...
+
+/* globals shared by all threads */
+z_stream c_stream;
+
+void* comp_file_multith (void *iterations)
+{
+        int err;
+        long *sleep_sec;
+
+	pid_t my_tid = syscall(SYS_gettid);
```
tuliom, line 19: Wrong indentation?  This error appears on other lines too.

test/test_pid_reuse.c
```diff
@@ -0,0 +1,122 @@
...
+void fork_children (int parent_pid) {
+	int child_id, i = 0;
+	int seconds = 10;
+	int children = 2048;
+
+	for (i=0; i < children; i++) {
+		child_id = fork();
+
+		if (child_id > 0) {
+			if (child_id == parent_pid) {
```
tuliom, line 45: Is this test targeting the kernel?
What is the relationship between fork() and libnxz?

test/test_pid_reuse.c
```diff
@@ -0,0 +1,122 @@
...
+				fprintf(stderr, "!!! pid %d reused (see ps)\n",
+					child_id);
+				sleep(10000);
```
tuliom, line 48: Why is this sleep necessary?
That's 10000 seconds.  Why not exiting?

test/test_pid_reuse.c
```diff
@@ -0,0 +1,122 @@
...
+			}
+			continue;
+		} else if (child_id == 0) {
+			pid_t pid = getpid();
+			fprintf(stderr, "%d: child(%d)\n", i, (int) pid);
+			sleep(seconds);
```
tuliom, line 54: Why is it sleeping 10 seconds?

test/test_pid_reuse.c
```diff
@@ -0,0 +1,122 @@
...
+			fprintf(stderr, "%d: child(%d) exits\n", i, (int) pid);
+			exit(0);
+		} else if (child_id == -1) {
+			fprintf(stderr, "fork failed: %s\n", strerror(errno));
+			sleep(2);
```
tuliom, line 59: Likewise...

test/test_pid_reuse.c
rzinsly: Yes, it is testing the kernel, what the test does is open an NX window and than fork other process to see if one of the child threads can use the same PID of the first one with the window still open.

I'll add more comments for the V2.

rzinsly: We could use pthread_create instead of fork but I don't think it matters.

test/test_pid_reuse.c
rzinsly: To keep this child alive until the point we get the csb error signal.

test/test_pid_reuse.c
rzinsly: To give time to all children be up and not reuse the same PID of a closed child.

test/test_pid_reuse.c
rzinsly: In fact here we could exit.

test/test_pid_reuse.c
rzinsly: Indeed, thanks for catching that.

rzinsly:
I rebased the patch with the current develop branch, fixed indentation, added more comments and updated the sleep for when the PID is reuse. The time intended was never 10000 but 1000 to be around 15 minutes, but in the system I tested 5 minutes is enough time to catch it.

test/test_pid_reuse.c
```diff
@@ -0,0 +1,166 @@
...
+	return (void *) 0;
+}
+
+void fork_children (int parent_pid) {
+	int child_id, i = 0;
+	int seconds = 10;
+	int children = 2048;
+
+	for (i=0; i < children; i++) {
+		child_id = fork();
```
pmur, line 76: Is this explicitly using a double-fork trick to avoid zombies?

test/test_pid_reuse.c
```diff
@@ -0,0 +1,166 @@
...
+/* Globals shared by all threads. */
+z_stream c_stream;
+
+void* comp_file_multith (void *iterations)
+{
+	int err;
+	long *sleep_sec;
+
+	pid_t my_tid = syscall(SYS_gettid);
+	fprintf(stderr, "thread started: pid: %d\n", (int) my_tid);
```
pmur, line 54: Should the message read tid?

test/test_pid_reuse.c
pmur: I think I am missing something subtle.  When can the forked child's pid ever be the parent's pid?

test/test_pid_reuse.c
```diff
@@ -0,0 +1,166 @@
...
+	long iterations;
+	z_const char hello[] = "hello, hello!";
+
+	if (argc == 2)
+		iterations = atoi(argv[1]);
+	else
+		iterations = 100;
+
+	/* Get parent thread PID. */
+	pid_t my_tid = syscall(SYS_gettid);
```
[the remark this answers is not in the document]
rzinsly: We are looking at the kernel PID, getpid returns the user PID that is the same for all threads in the same process.

test/test_pid_reuse.c
rzinsly: No, it is forking 2048 times to see if any child uses the same pid as the parent.

test/test_pid_reuse.c
rzinsly: It shouldn't, there is a bug in kernel that a thread reuses a PID of another thread that is not running but left a NX window open, this test is checking if the system has a kernel with this problem.

lucmaga: Is there already a fix for this kernel bug?

> It shouldn't, there is a bug in kernel that a thread reuses a PID of another thread that is not running but left a NX window open, this test is checking if the system has a kernel with this problem.

IMHO this explanation should be with the description on the file header. Also pointing 
out the kernel versions with this problem.

rzinsly: > Is there already a fix for this kernel bug?

Yes

test/test_pid_reuse.c
```diff
@@ -0,0 +1,167 @@
...
+	if (rc != 0) {
+		fprintf(stderr, "error: pthread cannot be joined %p"
+			"\n", ret);
+		return rc;
+	}
+
+	/* Fork many children to see if any use the same PID as the parent. */
+	fork_children((int) my_pid);
+	/* Wait for all children to run. */
+	sleep(60);
```
tuliom, line 147: If the goal is to wait for all children processes, you can use `wait()`, calling it in a loop until it returns -1 and `errno == ECHILD`.

rzinsly: Does it triggers when the child is sleeping?
Because we need to continue while the child that reused the PID is sleeping, not killed. (line 85)

tuliom: > Does it triggers when the child is sleeping?

@rzinsly No, it does not.

> we need to continue while the child that reused the PID is sleeping, not killed. (line 85)

In that case, I'm afraid there is a bug in this code.
The parent could exit before all its children complete without calling `wait()`, leaving zombie processes behind.

rzinsly: It shouldn't 60 seconds proved enough to wait for all children to finish or to get to the point that we hit the bug, in this case it will segfault later.
I added the wait() in loop at the end of the test for safety anyway.

test/test_pid_reuse.c
```diff
@@ -0,0 +1,172 @@
...
+	void *ret;
+	long iterations;
+	z_const char hello[] = "hello, hello!";
+
+	if (argc == 2)
+		iterations = atoi(argv[1]);
+	else
+		iterations = 100;
+
+	/* Get parent thread kernel PID. */
```
pmur, line 118: Should this read "Get parent thread ID"?  Similarly renaming the variable.

rzinsly: @pmur, we are looking for a kernel problem, I believe it's better to use the kernel naming and call it PID instead of TID.

PR #13, opened 2020-04-23
SetDictionary functions need testing before making them available

abalib:
Signed-off-by: Bulent Abali <abali@us.ibm.com>

### DIFF
--- a/lib/nx_deflate.c
+++ b/lib/nx_deflate.c
@@ -2163,7 +2163,8 @@ int nx_deflateSetDictionary(z_streamp strm, const unsigned char *dictionary, uns
 	uint32_t adler;
 	int cc;
 
-	//sw_trace("%s\n", __FUNCTION__);
+	return Z_STREAM_ERROR; /* TODO untested */
+
 	if (dictionary == NULL || strm == NULL)
 		return Z_STREAM_ERROR;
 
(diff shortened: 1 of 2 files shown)


abalib:
We didn't test these thoroughly.  We should disable them until then.  I propose to also apply the commit to the master branch as well

rzinsly:
I don't think it's a good idea remove useful functions and functionalities from the library, instead we should test it better and report bugs we encounter, also having it available we can get users data if they found issues.
The codes seems being working ok, have you found any problems?

abalib:
We need to run regression to make sure deflateSetDictonary output can be correctly decompressed.  I am still looking for the source data that failed.  The test is found here  https://github.com/libnxz/power-gzip/blob/develop/samples/dict-test.sh

abalib:
Thank you. I am going to setup some test runs to reproduce fails.  I assume it's ok to merge PR for now?

rzinsly:
> Thank you. I am going to setup some test runs to reproduce fails. I assume it's ok to merge PR for now?

Yes

abalib:
I will look for the test case.  But regardless I recognized later that my dictionary design and logic might be incorrect.   The worst thing we can do is giving users code that sometimes compresses incorrectly, and they cannot decompress their data later.  Until we fix this path, we should return an error.  Allow me to read the code and I come back with a better explanation and perhaps a design.

Issue #18, opened 2020-04-28
Sample code use internal functions and structures


tuliom:
The lack of a clear separation between internal and external API (PR #17) is affecting the sample code. The current code is making heavy usage of local functions.

This is clear when looking at [samples/Makefile](https://github.com/libnxz/power-gzip/blob/master/samples/Makefile), which has rules to rebuild files from `lib/`.

Issue #17, opened 2020-04-28
Headers are exposing internal structures, functions and macros


tuliom:
Ideally, libnxz would install headers that export only global functions and structures that are part of the ABI.
However, the current implementation makes it unclear to users and downstream what should really be exported.

This is particularly more complicated because the project does not provide a `make install` target yet (PR #11 ).

pedromfcarvalho:
There also seem to be certain global functions that have no declaration in the header, e.g. nx_adler32_z. Some of them are only accessible through the zlib wrappers, and not with the standalone library.

PR #19, opened 2020-04-29
A couple of small improvements

tuliom:
This is a couple of improvements and fixes to samples/Makefile and lib/nx_zlib.h that do not solve, but helps with issues #17 and #18 .

### DIFF
--- a/lib/nx_zlib.h
+++ b/lib/nx_zlib.h
@@ -89,9 +89,9 @@
 
 #define NXQWSZ  (sizeof(nx_qw_t))
 
-extern FILE *nx_gzip_log;
-extern FILE *nx_gzip_critical_log;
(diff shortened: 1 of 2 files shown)


tuliom:
@pedromfcarvalho Good point, but I don't think that's really important here.
Independently of the features, this code should be linking against libraries.

pedromfcarvalho:
> @pedromfcarvalho Good point, but I don't think that's really important here.
> Independently of the features, this code should be linking against libraries.

Ok. As far as I can tell the only real consequence is that lib/nx_dht might or might not print debug output independently from the debug output printed by the samples code (due to DHTPRT being guarded by NXDBG in lib/). The user would have to recompile the library to match the flags, but I don't think it's a big deal.

tuliom:
OK.  So, let me merge this.
Thanks!

PR #23, opened 2020-05-14
Tests with page faults

rzinsly:
This set of patches tests if the system can handle multiple page faults and if the kernel is disabling IRQ.

### DIFF
--- a/samples/bad_irq_check.c
+++ b/samples/bad_irq_check.c
@@ -1,0 +1,46 @@
+/*
+ * Wrappers to skip touching pages.
+ * We force a page fault to see if the kernel is disabling the VAS IRQ, keeping
+ * the NX windows open and preventing other process to use NX anymore.
+ *
+ * Copyright (C) 2020 IBM Corporation
+ *
+ * Licenses for GPLv2 and Apache v2.0:
+ *
+ * GPLv2:
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * Apache v2.0:
+ *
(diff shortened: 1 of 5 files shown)


lucmaga:
LGTM.
It might be good to add a README file here and warn about the unsafe tests.

samples/bad_irq_check.c
```diff
@@ -0,0 +1,44 @@
+/*
+ * Wrappers to skip touching pages and test if the kernel is disabling IRQ.
```
pmur, line 2: It would help my understanding to have more detail about the motivation behind this test inlined here too.  The commit description still seems somewhat terse IMO.

samples/rand_pfault_check.c
```diff
@@ -0,0 +1,58 @@
+/*
+ * Wrappers to randomly touch pages.
```
pmur, line 2: Extra description here would also help me to understand the motivations behind this patch, I would also recommend embedding some form of the commit message here too.

rzinsly:
> LGTM.
> It might be good to add a README file here and warn about the unsafe tests.

Our README file still don't mention the samples folder, I plan to update that, as well as the wiki page.

samples/run-series.sh
```diff
@@ -39,8 +39,17 @@ for i in 1 4 16 32 64 80; do
...
     NX_GZIP_PASTE_RETRIES=200 ./compdecomp_th junk2 400 24
+    # Stress test for disabling IRQ on kernel.
+    echo "Testing for kernel disabling IRQ..."
+    NX_GZIP_TIMEOUT_PGFAULTS=3 ./bad_irq_check junk2 400 96
+    echo "Success!"
+    # Stress test to check if the system handles many page faults.
+    echo "Checking many random page faults..."
+    ./rand_pfault_check junk2 100 480
+    echo "Success!"
 fi
```
tuliom, line 55: Everything in this unsafe section are tests that are saved in the samples directory.
Why are we storing tests under samples/?  Should they be stored under test/?

rzinsly: They are under samples/ because they depend on compdecomp_th. We could move they to tests/ but IMO it's better not have files in tests/ depending on files on samples/.
With some work on samples we may be able to move compdecomp_th to tests, I'm not sure run-series.sh should be on samples/ but this is in a larger scope than this PR.

tuliom: > but this is in a larger scope than this PR.
Agreed.
Could you create an issue for this so that we don't miss it, please?

PR #26, opened 2020-06-16
Create an installable header

### LINKED ISSUE #17: Headers are exposing internal structures, functions and macros

rzinsly:
Adds the libnxz.h header containing all exported functions without
exposing internal symbols, this fixes #17.

### DIFF
--- a/lib/Versions
+++ b/lib/Versions
@@ -6,6 +6,7 @@ LIBNXZ_0.61.0 {
     nx_deflateSetHeader;
     nx_inflateGetHeader;
     nx_inflateResetKeep;
+    nx_deflateResetKeep;
     nx_uncompress2;
     nx_inflateInit_;
     nx_inflateInit2_;
--- a/libnxz.h
+++ b/libnxz.h
@@ -1,0 +1,139 @@
+/*
+ * NX-GZIP compression accelerator user library
+ * implementing zlib library interfaces
+ *
+ * Copyright (C) IBM Corporation, 2020
+ *
+ * Licenses for GPLv2 and Apache v2.0:
+ *
+ * GPLv2:
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ *
+ * Apache v2.0:
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#include <stdint.h>
+#include <stddef.h>
+
+#ifndef _LIBNXZ_H
+#define _LIBNXZ_H
+
+/* zlib crc32.c and adler32.c */
+extern ulong nx_adler32(ulong adler, const char *buf, uint len);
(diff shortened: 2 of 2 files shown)


lib/Versions
```diff
@@ -29,6 +30,7 @@ LIBNXZ_0.61.0 {
     nx_crc32;
     nx_crc32_combine;
     nx_crc32_combine64;
+    crc32_ppc;
```
tuliom, line 33: Why is this a global symbol?

libnxz.h
```diff
@@ -0,0 +1,147 @@
...
+#endif
+
+/* zlib crc32.c, crc32_ppc.c and adler32.c */
+extern ulong nx_adler32(ulong adler, const char *buf, uint len);
+extern ulong nx_adler32_z(ulong adler, const char *buf, size_t len);
+extern ulong nx_adler32_combine(ulong adler1, ulong adler2, uint64_t len2);
+extern ulong nx_crc32(ulong crc, const unsigned char *buf, uint64_t len);
+extern ulong nx_crc32_combine(ulong crc1, ulong crc2, uint64_t len2);
+extern ulong nx_crc32_combine64(ulong crc1, ulong crc2, uint64_t len2);
+extern ulong crc32_ppc(ulong crc, unsigned char *buffer, unsigned len);
```
tuliom, line 54: I don't understand why this symbol is listed here.  Aren't `crc32` and `nx_crc32` enough?

libnxz.h
```diff
@@ -0,0 +1,147 @@
...
+extern ulong nx_compressBound(ulong sourceLen);
+
+/* nx_uncompr.c */
+extern int nx_uncompress2(unsigned char *dest, ulong *destLen,
+			  const unsigned char *source, ulong *sourceLen);
+extern int nx_uncompress(unsigned char *dest, ulong *destLen,
+			 const unsigned char *source, ulong sourceLen);
+
+
+#ifdef ZLIB_API
```
tuliom, line 105: This is exposing an internal macro to the user.
Is this really intentional?
Can't we export these functions without depending on this macro?

libnxz.h
```diff
@@ -0,0 +1,147 @@
...
+extern ulong nx_crc32_combine64(ulong crc1, ulong crc2, uint64_t len2);
+extern ulong crc32_ppc(ulong crc, unsigned char *buffer, unsigned len);
+
+/* nx_deflate.c */
+extern int nx_deflateInit_(void *strm, int level, const char *version,
+			   int stream_size);
+extern int nx_deflateInit2_(void *strm, int level, int method, int windowBits,
+			    int memLevel __unused, int strategy,
+			    const char *version __unused,
+			    int stream_size __unused);
```
tuliom, line 62: I don't think it's necessary to use `__unused` here.

lib/Versions
rzinsly: To allow users to use the crc32 optimized version if ZLIB_API is not defined.

libnxz.h
rzinsly: crc32 calls crc32_ppc, but is not available if ZLIB_API isn't defined.
But I got your point, maybe nx_crc32 should call crc32_ppc as well.

libnxz.h
rzinsly: Was intentional because we don't need these functions if the macro is not defined, but yes we can export them without it, I'll change that.

libnxz.h
tuliom: Just to be clear: I don't think you have to implement it for this patch.
But I don't think we should add an extra symbol.

By the way, I don't understand why we need nx_crc32 with a generic implementation.
This library only works on P9 and the crc32_ppc code runs on P9.  In other words: we don't even need an IFUNC implementation.

tuliom: I've just had a chat with Bulent and he explained to me the header checksum is just a few bytes.  In this scenario, I wouldn't be surprised if the generic implementation is faster than the vectorized one.
He also pointed out that we might not need the fastest implementation because the length is too small to make a difference.

rzinsly: @tuliom I got it, thanks!
Just removed this addition.

libnxz.h
```diff
@@ -0,0 +1,139 @@
...
+ *
+ */
+
+#include <stdint.h>
+#include <stddef.h>
+
+#ifndef _LIBNXZ_H
+#define _LIBNXZ_H
+
+/* zlib crc32.c and adler32.c */
```
lucmaga, line 43: Just a small question here.
Are this all the zlib symbols exported by libnxz?

rzinsly: I don't believe libnxz has all implemented, but most of them are.

Issue #27, opened 2020-06-16
Build issue with 'make bench'


tuliom:
I started to see this error:

```
/usr/bin/gcc -std=gnu11 -O3 -fPIC -D_LARGEFILE64_SOURCE=1 -DHAVE_HIDDEN -mcpu=power9 -DZLIB_API  -o gunzip_nx_test gunzip_nx.o ../lib/libnxz.a -lpthread
../lib/libnxz.a(nx_zlib.o):(.bss+0x4020): multiple definition of `nx_gzip_log'
gunzip_nx.o:(.bss+0x0): first defined here
../lib/libnxz.a(nx_zlib.o):(.bss+0x402c): multiple definition of `nx_dbg'
gunzip_nx.o:(.bss+0x8): first defined here
../lib/libnxz.a(nx_zlib.o): In function `sigsegv_handler':
nx_zlib.c:(.text+0x4bd0): multiple definition of `sigsegv_handler'
gunzip_nx.o:gunzip_nx.c:(.text+0x0): first defined here
collect2: error: ld returned 1 exit status
make[1]: *** [Makefile:31: gunzip_nx_test] Error 1
```

git-bisect is pointing to the following commit:

```
commit 4146ff949a579f9a7ecc6daeeb564b2b8ad13bbf
Merge: be87cbb e643339
Author: Raphael Moreira Zinsly <6718397+rzinsly@users.noreply.github.com>
Date:   Thu Jun 4 09:38:09 2020 -0300

    Merge pull request #23 from rzinsly/system_tests
    
    Tests with page faults
```

tuliom:
Fixed in PR #28.

PR #28, opened 2020-06-16
Fix build issues

tuliom:
Fix issue #27 and other small issues.
One of the patches is only removing dead code while the other one avoids running an unsafe test while running `make check`

### DIFF
--- a/lib/nx_zlib.c
+++ b/lib/nx_zlib.c
@@ -89,7 +89,6 @@ pthread_mutex_t nx_devices_mutex; /* mutex to protect global stats */
 struct zlib_stats zlib_stats;	/* global statistics */
 
 struct sigaction act;
-void sigsegv_handler(int sig, siginfo_t *info, void *ctx);
 /* **************************************************************** */
 
 /*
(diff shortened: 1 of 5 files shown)


test/Makefile
```diff
@@ -17,7 +17,7 @@ obj_test_pid_reuse = ${patsubst %.c, %.o, test_pid_reuse.c}
 
 TEST_OBJS = $(sort $(obj_test_deflate) $(obj_test_inflate) $(obj_test_stress) \
     $(obj_test_crc32) $(obj_test_adler32) $(obj_test_multithread_stress))
-EXE = test_inflate test_deflate test_stress test_crc32 test_adler32 \
```
rzinsly, line 20: If you are removing it here it should be removed from run_tests.sh as well.
Did you had any problem with test_stress? It shouldn't be unsafe unless you drastic reduce the max poll number for wait_for_csb().

tuliom: > Did you had any problem with test_stress?

Yes, when running `make -j$(nproc) check unsafe-check`.
But let me drop this patch for now.  I'll investigate this issue later...

tuliom:
@rzinsly After removing one of the commits, this is now just removing dead code.

Issue #11, opened 2020-04-15
Provide the install target


tuliom:
Most projects provide an `install` target that helps downstream with the installation process.
It helps downstream know which files need to be installed, e.g. *.so, *.a, documentation and header files.

Without that, downstream is forced to find these files by itself, at the risk of missing an important one.

rzinsly:
Done by https://github.com/libnxz/power-gzip/pull/31

PR #31, opened 2020-06-23
Adds install/uninstall target

rzinsly:
Adds the install and uninstall targets to Makefile in order to facilitate
the installation of the library.
Solves #11.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+include config.mk
 
 subdirs = lib test samples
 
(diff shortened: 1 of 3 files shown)


mscastanho:
@rzinsly Overall LGTM. Just curious: why does the shared lib need to be executable?

tuliom:
> @rzinsly LGTM. Just curious: why does the shared lib need to be executable?

@mscastanho Because shared libraries are executed.

Issue #38, opened 2020-07-20
Stop using `ERR_NX_TRANSLATION`


tuliom:
A bug has been fixed in the kernel where the return error after an address translation failure.
With that said, it's required to stop using `ERR_NX_TRANSLATION` and use a new value instead (250).

We may also need to add some temporary code that is able to deal with both error codes until all kernels in use are fixed.

Reference: https://lists.ozlabs.org/pipermail/linuxppc-dev/2020-July/214190.html

PR #39, opened 2020-08-04
Use proper error code to check fault address

rzinsly:
The completion codes were fixed on kernel (#38), now ERR_NX_TRANSLATION
(CSB.CC=5) is only used for internal VAS faults handling and should
not used by libnxz. ERR_NX_AT_FAULT (CSB.CC=250) is the proper error
code reported by the kernel when NX encounters address translation
failure.

### DIFF
--- a/lib/nx_deflate.c
+++ b/lib/nx_deflate.c
@@ -1476,8 +1476,8 @@ restart:
 	}
 
 	switch (cc) {
-	case ERR_NX_TRANSLATION:
-		prt_warn("ERR_NX_TRANSLATION: bytes_in %d nxcmdp->crb.csb.fsaddr %p\n",
+	case ERR_NX_AT_FAULT:
+		prt_warn("ERR_NX_AT_FAULT: bytes_in %d nxcmdp->crb.csb.fsaddr %p\n",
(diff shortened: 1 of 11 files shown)


rzinsly:
> Maybe you can remove the definition of `ERR_NX_TRANSLATION` if its not being used anywhere else.

We have all possible error codes defined, maybe we can remove the ones that we don't use but removing only `ERR_NX_TRANSLATION` seems wrong.

tuliom:
It's my understanding this modification will break compatibility with older versions of the kernel patch.
Is there a way to make the change and continue to support older kernels?

rzinsly:
> It's my understanding this modification will break compatibility with older versions of the kernel patch.
> Is there a way to make the change and continue to support older kernels?

I tested with an older kernel and works fine.
This doesn't actually break anything, the patch changes error handling, if the compression goes well nothing changes.
Applying this patch with older kernel versions if we got an address translation fault the application will stop and print the CC value that caused it instead of printing an warning, treating the error and retrying with fewer pages.
The only way I can think of keeping the same behaviour on both older and new kernel versions is by treating CC 5 and CC 250 the same, but if we got the internal CC 5 after the kernel fix something went wrong and for me it seems we should stop the application.

Issue #56, opened 2021-03-16
Fix all `-Wpointer-sign` warnings


tuliom:
There are 314 warnings in `make check` pointed out by `-Wpointer-sign`.

PR #57, opened 2021-03-16
Update the documentation and enable -Wall -Werror

tuliom:
This PR has 1 patch to add more documentation, 5 patches fixing warnings pointed out by the compiler and a last patch to enable `-Wall -Werror`.

Notice that, because of issue #56 , there are still warnings pending.


tuliom:
This PR has been opened against the wrong branch.  See PR #58 for the correct one.

Issue #59, opened 2021-03-17
Tests should abort in case of error


tuliom:
test_deflate and test_inflate are not treating all kinds of errors, e.g. `Z_STREAM_ERROR`.

In the exceptional scenario that one of these errors appear, they're stuck in an infinite loop.

I found this error in a scenario where the OS returned `cc=250`.

tuliom:
Fixed with PR #60 .

PR #60, opened 2021-03-19
Fix for issue #59 and improving issue #56

tuliom:
The first patch fixes issue #59 , but also deduplicates code in the tests.

The second patch reduces the amount of warnings described in issue #56 .

Suggestion: you might want to review the patches separately because the modify a lot of code, but they're very simple.


rzinsly (approved):
LGTM but I would prefer more meaningful names on the z_finish and test_stress functions to better differentiate, e.g. `_test_nx_deflate_finish`, `_test_deflate_stress`, etc.

PR #72, opened 2021-09-03
Addapt nx_enumerate_engines() to work with PowerVM

### LINKED ISSUE #62: Change nx_dev_count on PowerVM
On PowerVM, there is no concept of multiple NX engines.
The following code is known to work on PowerVM, but is definitely unacceptable as a generic solution:
```
diff --git a/lib/nx_zlib.c b/lib/nx_zlib.c
index c4c9dec..dbc07ed 100644
--- a/lib/nx_zlib.c
+++ b/lib/nx_zlib.c
@@ -997,13 +932,13 @@ void nx_hw_init(void)
                }
        }
 
-       nx_count = nx_enumerate_engines();
+       /*nx_count = nx_enumerate_engines();
        nx_dev_count = nx_count;
        if (nx_count == 0) {
                prt_err("NX-gzip accelerators found: %d\n", nx_count);
                return;
-       }
-
+               }*/
+       nx_dev_count = 1;
        prt_info("%d NX GZIP Accelerator Found!\n",nx_count);
 
        if (trace_s != NULL)
```

rzinsly:
This patch detects the platform where libnxz is being built
and sets 1 NX engine for PowerVM on nx_enumerate_engines().
Fixes #62.

Signed-off-by: Raphael Moreira Zinsly <rzinsly@linux.ibm.com>

### DIFF
--- a/lib/nx_zlib.c
+++ b/lib/nx_zlib.c
@@ -665,6 +665,11 @@ static int nx_enumerate_engines()
 			count++;
 
 		}
+		/* On PowerVM, there is no concept of multiple NX engines.  */
+		if (strncmp(de->d_name, "ibm,powervm", 11) == 0){
+			closedir(d);
+			return 1;
+		}
 	}
 
 	closedir(d);


tuliom (changes requested):
I think the test code is very close to what we need, but I think that `platform(s)` must not be evaluated at build time because this will force downstream to provide 2 different libnxz builds in order to support both PowerNV and pSeries.

tuliom (approved):

PR #82, opened 2021-10-07
Remove unnecessary checks during configuration processing

mscastanho:
GCC 12 (dev) started to complain that some NULL checks were not necessary as the
values could never be NULL. The warnings seem to be correct, so the checks have
been removed.

In find_cfg and nx_dump_cfg the loop iterates over a maximum of
cfg_table->cfg_num elements. As long as cfg_num is at most MAX_CONFIG_ITEM, we
are fine because cfg_table can hold that many elements, and key and val are
pointers to "static" arrays inside struct cfg_item.

In set_cfg, we access key and val of item at index cfg_cnt, which is guarded by
the function to be less than MAX_CONFIG_ITEM, so we are guaranteed to access
initialized data.

Tested with gcc 11.1.1 and gcc 12.0.0 (rev 5f0285418940).

### DIFF
--- a/lib/nx_utils.c
+++ b/lib/nx_utils.c
@@ -99,8 +99,6 @@ static int find_cfg(char *key, struct nx_cfg_tab *cfg_table)
 
 	configs = cfg_table->configs;
 	for (i = 0; i < cfg_table->cfg_num; i++) {
-		if (!configs[i].key)
-			continue;
 		if (strcmp(key, configs[i].key) == 0)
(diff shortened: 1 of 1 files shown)


rzinsly (approved):
I guess the only we a key could be `NULL` is if you read a conf file with a literally `NULL` key, but I don't think we should treat that.
LGTM

tuliom (approved):

PR #85, opened 2021-10-15
Fix valgrind issues on test_multithread_stress and add valgrind-check target

lucmaga:


test/valgrind.supp
```diff
@@ -3,3 +3,10 @@
   Memcheck:Cond
   fun:compare_data
 }
+{
+  Handle pthread/valgrind issue (possible leak)
+  Memcheck:Leak
+  fun:calloc
+  fun:calloc
+  fun:allocate_dtv
+}
```
tuliom, line 12: @lamm Why is this suppression needed?
Could you mention this in the commit message, please?

test/Makefile
```diff
@@ -23,9 +23,17 @@ UNSAFE = test_stress
 
 all:
 
-check: $(EXE)
+check: $(EXE) plain-check valgrind-check
```
tuliom, line 26: I don't think it's a good idea to require that a `make check` to be executed on Valgrind.
The support for copy and paste was added to Valgrind recently limiting where tests could be executed.
Even if it was added long ago, I don't think manual tests should be forcing tests on top of Valgrind.

test/Makefile
```diff
@@ -26,6 +26,12 @@ all:
 check: $(EXE)
 	./run_test.sh
 
+valgrind-check: $(EXE)
+ ifeq (, $(shell which valgrind))
+ $(error "No valgrind found in $(PATH), skipping valgrind tests")
+ endif
+	TEST_WRAPPER="valgrind --leak-check=full --suppressions=valgrind.supp" ./run_test.sh
```
tuliom, line 33: I was thinking more about this and I'm afraid this is going to cause trouble in the future, e.g. if we decide to stop running tests serially and start running them in parallel, this recipe would have to be rewritten.
I believe we can still provide a `valgrind-check` target in the main Makefile (this patch already does that), but running:
`$(MAKE) -C test check TEST_WRAPPER="valgrind --leak-check=full --suppressions=valgrind.supp"`

I also suggest to remove the test for valgrind in this recipe.

test/valgrind.supp
lucmaga: Actually, the issue was that now all the threads were being joined.
Commit 39ad027 now fix this.

test/test_multithread_stress.c
```diff
@@ -222,7 +224,10 @@ int main(int argc, char **argv)
 	printf("Test Interval  :\t%d\n", test_interval);
 	printf("Test Iterations:\t%d\n",test_iterations);
 
-	generate_data_buffer(data_buf);
+	if (generate_data_buffer(data_buf) != 0) {
+		free_data_buffer(data_buf);
```
mscastanho, line 213: Is this 100% safe? If the OOM happens very early in `generate_data_buffer`, it will exit without initializing some elements in `buffer`, not even with `NULL`. So `free_data_buffer` may call `free` on uninitialized pointers later. The man page for `free` does not make it really clear what happens in this case.

test/test_multithread_stress.c
```diff
@@ -229,6 +229,9 @@ int main(int argc, char **argv)
 		/*Check for finish*/
 		for (i = 0; i < thread_num; i++){
 			if (!thread_info[i].running && (thread_info[i].iteration < test_iterations)) {
+				/*  Make sure to join the thread before reuse the thread_info field. Otherwise
+				   the internal thread resources will leak. */
```
mscastanho, line 233: Small nit: align text with previous line.

Makefile
```diff
@@ -16,6 +16,9 @@ bench: $(subdirs)
 check:  $(subdirs)
 	$(MAKE) -C test $@
 
+valgrind-check:  $(subdirs)
+	$(MAKE) -C test check TEST_WRAPPER="valgrind --leak-check=full --suppressions=valgrind.supp"
```
mscastanho, line 20: I think `--error-exitcode=1` could also be helpful here, to guarantee valgrind exits with a non-zero error code when it finds something odd.

test/test_multithread_stress.c
lucmaga: Actually this is just fixed if the `data_buf` is initialized with `NULL`.

test/test_multithread_stress.c
```diff
@@ -229,6 +229,9 @@ int main(int argc, char **argv)
 		/*Check for finish*/
 		for (i = 0; i < thread_num; i++){
 			if (!thread_info[i].running && (thread_info[i].iteration < test_iterations)) {
+				/*  Make sure to join the thread before reuse the thread_info field.
+				    Otherwise the internal thread resources will leak. */
+				pthread_join(thread_info[i].tid, NULL);
```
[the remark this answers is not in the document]
rzinsly: I think it impacts, this test was created to simulate a database workload.
I'm no very familiar with this kind of workload but I believe it should have the "thread-bomb" behavior.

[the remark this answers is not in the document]
lucmaga: @tuliom, @rzinsly IMO the impact should be minimal as `thread_info[i].running` is 0 already.

@rzinsly Other option is to change it to `pthread_tryjoin_np` and just continue the loop if we can't join the thread yet.

PR #88, opened 2021-10-15
Make soft_copy_threshold configurable and debuggable

mscastanho:
- Let user adjust soft_copy_threshold
- Print current soft_copy_threshold value on verbose mode


lucmaga (approved):

mscastanho:
@tuliom I added the restriction for minimum and maximum values, and updated the commit message to document a bit how those values were chosen.

lib/nx_zlib.c
```diff
@@ -1088,6 +1092,12 @@ void nx_hw_init(void)
 		nx_config.timeout_pgfaults = str_to_num(timeout_pgfaults);
 	}
 
+	if (soft_copy_threshold) {
+		nx_config.soft_copy_threshold = str_to_num(soft_copy_threshold);
+		nx_config.soft_copy_threshold = NX_MAX(nx_config.soft_copy_threshold, 31);
+		nx_config.soft_copy_threshold = NX_MIN(nx_config.soft_copy_threshold, 8190);
```
rzinsly, line 1099: Do we know which value is usually the best? If so we should default to it, otherwise this patches LGTM.

mscastanho: @rzinsly, @tuliom has been doing some investigations on this threshold, but I don't think we have a "best" value so far. Right now if the user doesn't specify anything, it will keep using 1024 as the default like before.

Issue #45, opened 2020-11-17
Feature: Fallback to software zlib in case NX is not available


carlosedp:
It would be a great addition to the library to add the software zlib implementation as a fallback mechanism in case there is no NX in the system or no NX instances available in time of processing.

PR #93, opened 2021-10-21
Add an automatic switch mode to gzip_selector

rzinsly:
Adds a mode that automatically switches between software and hardware
compression depending on the input size and make it the default mode.

### DIFF
--- a/lib/nx_uncompr.c
+++ b/lib/nx_uncompr.c
@@ -77,8 +77,11 @@ int uncompress2(Bytef *dest, uLongf *destLen, const Bytef *source, uLong *source
 {
 	int rc;
 
-	if(nx_config.gzip_selector == GZIP_MIX){
-		rc = sw_uncompress2(dest, destLen, source, sourceLen);
+	if(nx_config.gzip_selector == GZIP_AUTO){
+		if(*sourceLen <= DECOMPRESS_THRESHOLD)
+			rc = sw_uncompress2(dest, destLen, source, sourceLen);
+		else
+			rc = nx_uncompress2(dest, destLen, source, sourceLen);
 	}else if(nx_config.gzip_selector == GZIP_NX){
 		rc = nx_uncompress2(dest, destLen, source, sourceLen);
 	}else{
@@ -99,8 +102,11 @@ int uncompress(Bytef *dest, uLongf *destLen, const Bytef *source, uLong sourceLe
 #else
 	int rc=0;
 
-	if(nx_config.gzip_selector == GZIP_MIX){
-		rc = sw_uncompress(dest, destLen, source, sourceLen);
+	if(nx_config.gzip_selector == GZIP_AUTO){
+		if(sourceLen <= DECOMPRESS_THRESHOLD)
+			rc = sw_uncompress(dest, destLen, source, sourceLen);
+		else
+			rc = nx_uncompress(dest, destLen, source, sourceLen);
 	}else if(nx_config.gzip_selector == GZIP_NX){
 		rc = nx_uncompress(dest, destLen, source, sourceLen);
 	}else{
--- a/lib/nx_zlib.c
+++ b/lib/nx_zlib.c
@@ -971,7 +971,7 @@ void nx_hw_init(void)
 	nx_config.dht = 0; /* default is literals only */
 	nx_config.nx_ratio = 100; /* default is 100% NX */
 	nx_config.strategy_override = 1; /* default is dynamic huffman */
-	nx_config.gzip_selector = GZIP_NX; /* default is to use NX only */
+	nx_config.gzip_selector = GZIP_AUTO; /* default to the automatic switch */
 
 	if (!cfg_file_s)
 		cfg_file_s = "./nx-zlib.conf";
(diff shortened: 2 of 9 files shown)


lib/nx_uncompr.c
```diff
@@ -99,8 +102,11 @@ int uncompress(Bytef *dest, uLongf *destLen, const Bytef *source, uLong sourceLe
 #else
 	int rc=0;
 
-	if(nx_config.gzip_selector == GZIP_MIX){
-		rc = sw_uncompress(dest, destLen, source, sourceLen);
```
mscastanho, line 103: This may be unrelated to this PR, but for `GZIP_MIX`, don't we also need to choose based on `nx_ratio` here? In this case `GZIP_MIX` and `GZIP_MIX2` are the same thing.

lib/nx_zlib.c
```diff
@@ -1025,10 +1025,11 @@ void nx_hw_init(void)
...
 	if(type_selector != NULL){
 		nx_config.gzip_selector = str_to_num(type_selector);
-		prt("gzip_selector: %d (1-SW;2-NX;3/4-MIX)\n", nx_config.gzip_selector);
-		if(nx_config.gzip_selector == 0 || nx_config.gzip_selector > 4) {
-			prt("Unrecognized option, defaulting to NX.\n");
-			nx_config.gzip_selector = GZIP_NX;
+		prt("gzip_selector: %d (0-AUTO;1-SW;2-NX;3/4-MIX)\n", nx_config.gzip_selector);
+		/* check for misconfiguration */
+		if(nx_config.gzip_selector > 4) {
+			prt("Unrecognized option, defaulting to AUTO.\n");
```
mscastanho, line 1031: These lines were only added in the second commit, but should actually be part of the first one, which sets AUTO as the default.

mscastanho (commented):
A few comments inline. Also, I'm a bit intrigued by the message in the second commit:

>We should not default to zlib if we don't have NX available

 Why not? I thought this was a planned feature given #45 is still open.

mscastanho:
Also, we should really add some tests to validate the different modes, as our current testsuite will only check the default mode. But I don't see a problem if they are introduced by a later PR.

lib/nx_zlib.h
```diff
@@ -332,6 +327,8 @@ static inline int use_nx_inflate(z_streamp strm)
 	/* #1 Threshold */
 	if(strm->avail_in <= DECOMPRESS_THRESHOLD) return 0;
 
+	if(nx_config.gzip_selector == GZIP_AUTO) return 1;
+
```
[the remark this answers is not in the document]
rzinsly: >The way the code is written now, we'll use software zlib every time the input is too small

No, we already test for the other modes before testing the threshold.

>Note that if the user has selected GZIP_MIX, and the input is small, software zlib will be used 100% of the time, regardless of nx_ratio.

That is true, but this always been like that for `GZIP_MIX`, by design it wouldn't use the ratio for small inputs because software is a better choice in this case. I think this is a good approach and I will update the `GZIP_MIX` description to reflect that, but if you guys prefer to only rely on the ratio for that mode I'm open to discuss it.

lib/nx_zlib.c
rzinsly: Agreed, I may have splitted this by mistake in the many rebases I did.

lib/nx_uncompr.c
rzinsly: This is another problem with the description. The reasoning behind `GZIP_MIX` decompression is the same as `GZIP_MIX2`, as the decompression throughput is usually higher on software there is not much reason to use the ratio for decompression as well, but I'm not against it.

lib/nx_zlib.h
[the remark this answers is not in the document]
mscastanho: @rzinsly I think making the description more clear is enough. Thanks!

rzinsly:
> Why not? I thought this was a planned feature given #45 is still open.

I forgot about that issue, but I disagree with that if the system doesn't have NX available IMHO the user shouldn't be using libnxz, but I think the opposite makes sense, if the system doesn't have zlib installed than the fallback is to use only NX.

lib/nx_uncompr.c
mscastanho: Fair enough. As I said in the other comment, improving the description of `GZIP_MIX` is enough IMO.

mscastanho (approved):

tuliom (commented):
I have 1 question.
Otherwise, LGTM.
We might need a way to switch to sw if the system NX are under high usage, but we can't implement that now as it would require to collect more data from the kernel.

lib/nx_zlib.c
```diff
@@ -1038,7 +1038,7 @@ void nx_hw_init(void)
 	if (nx_count == 0) {
 		nx_close_cfg(&cfg_tab);
 		prt_err("NX-gzip accelerators found: %d\n", nx_count);
-		nx_config.gzip_selector = GZIP_SW; /*fallback to use software zlib*/
+		return;
```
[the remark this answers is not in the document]
tuliom: I've just seen the other messages and I have an answer.

tuliom:
> if the system doesn't have NX available IMHO the user shouldn't be using libnxz.

Good point. This is a good way of thinking for a library that is meant to be used in specific scenarios.
However, it isn't a good behavior for a generically used library, i.e. one that is broadly used by programs.

I do agree that libnxz falls in the first category now.
However, I would love to see this code (doesn't need to be libnxz) being used in a generic way.

tuliom (approved):

PR #96, opened 2021-10-29
Add configuration cache_threshold

tuliom:
cache_threshold indicates the amount of data that should be cached
from nx_deflate() and nx_inflate() before sending a request to NX.

This new configuration replaces compress_threshold that was used
only in nx_deflate().

After this change, soft_copy_threshold is used only to indicate if the
copy/paste facility should be used or if memcpy() is preferred.

### DIFF
--- a/lib/nx_zlib.c
+++ b/lib/nx_zlib.c
@@ -875,6 +875,7 @@ static int print_nx_config(FILE *fp)
 	fprintf(fp, "dis_savedevp: %d\n", disable_saved_nx_devp);
 	fprintf(fp, "timeout_pgfaults: %d\n", nx_config.timeout_pgfaults);
 	fprintf(fp, "soft_copy_threshold: %d\n", nx_config.soft_copy_threshold);
+	fprintf(fp, "cache_threshold: %d\n", nx_config.cache_threshold);
 
 	return 0;
 };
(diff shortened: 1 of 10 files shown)


lib/nx_zlib.c
```diff
@@ -932,7 +934,8 @@ void nx_hw_init(void)
 	nx_config.per_job_len = (1024 * 1024); /* less than suspend limit */
 	nx_config.strm_def_bufsz = (1024 * 1024); /* affect the deflate fifo_out */
 	nx_config.soft_copy_threshold = 1024; /* choose memcpy or hwcopy */
-	nx_config.compress_threshold = (10*1024); /* collect as much input */
+	nx_config.cache_threshold = 8 * 1024; /* Cache input before
+					       *  processing */
```
lucmaga, line 938: Why change to 8KB instead of 10KB?

lib/nx_zlib.h
```diff
@@ -123,7 +123,7 @@ struct nx_config_t {
 	uint32_t per_job_len;          /* less than suspend limit */
 	uint32_t strm_def_bufsz;
 	uint32_t soft_copy_threshold;  /* choose memcpy or hwcopy */
-	uint32_t compress_threshold;   /* collect as much input */
+	uint32_t cache_threshold; /** Cache input before processing */
```
lucmaga, line 126: There are two '*' instead of one on the comment.

lucmaga (approved):
LGTM with the comment fixed.

rzinsly (approved):

lib/nx_zlib.h
tuliom: This is intentional.  That means it will appear in the Doxygen documentation.

lib/nx_zlib.c
tuliom: Ideally, we want to cache as little as possible and provide the highest throughput.
In my tests, caching 8KiB or 10KiB had the same performance.

tuliom:
I updated the patch in order to fix a conflict with the latest code in `develop`.

PR #100, opened 2021-11-09
Fix output when deflate is called with no input data (#76)

### LINKED ISSUE #76: When compressing 0 byte file, nxgzip and gzip command have different results.
When compressing 0 byte file, nxgzip result is 0 byte file. when using gzip command, the result is 29 bytes. 

[root@localhost tmp]# touch zerobyte
[root@localhost tmp]# LD_PRELOAD=./libnxz.so ./gzip_test zerobyte -c > zerobyte.nxgzip.gz
[root@localhost tmp]# gzip zerobyte -c > zerobyte.soft.gz
[root@localhost tmp]# ls -l
total 304
-rwxr-xr-x 1 root root  76712 Sep 17 17:27 gzip_test
lrwxrwxrwx 1 root root     14 Sep 17 17:28 libnxz.so -> libnxz.so.0.62
-rw-r--r-- 1 root root 227056 Sep 17 17:27 libnxz.so.0.62
-rw-r--r-- 1 root root      0 Sep 17 17:28 zerobyte
-rw-r--r-- 1 root root      0 Sep 17 17:29 zerobyte.nxgzip.gz
-rw-r--r-- 1 root root     29 Sep 17 17:29 zerobyte.soft.gz

mscastanho:
The issue that was causing #76 was a combination of early stream end check and special handling for calls with no input, please see the individual commits for more details.

Commit `Indent zpipe-issue99.c correctly` is here only for review purposes,
to facilitate checking the differences compared to `samples/zpipe.c`.

The output of `test_zeroinput` shows that now libnxz emits a non-empty output file even if the input stream is empty, but the generated deflate stream is different from zlib's.

```
~/src/power-gzip> LD_LIBRARY_PATH=./lib ./test/test_zeroinput 
-> empty data stream
Output for Z_NO_FLUSH: 0x78 0x9c 0x01 0x00 0x00 0xff 0xff 0x00 0x00 0x00 0x01
Output for Z_PARTIAL_FLUSH: 0x78 0x9c 0x01 0x00 0x00 0xff 0xff 0x00 0x00 0x00 0x01
Output for Z_SYNC_FLUSH: 0x78 0x9c 0x01 0x00 0x00 0xff 0xff 0x00 0x00 0x00 0x01
Output for Z_FULL_FLUSH: 0x78 0x9c 0x01 0x00 0x00 0xff 0xff 0x00 0x00 0x00 0x01
Output for Z_FINISH: 0x78 0x9c 0x01 0x00 0x00 0xff 0xff 0x00 0x00 0x00 0x01
-> mid-decompression zero input
*** test_zeroinput.c passed

~/src/power-gzip> LD_PRELOAD=/lib64/libz.so LD_LIBRARY_PATH=./lib ./test/test_zeroinput 
-> empty data stream
Output for Z_NO_FLUSH: 0x78 0x9c 0x03 0x00 0x00 0x00 0x00 0x01
Output for Z_PARTIAL_FLUSH: 0x78 0x9c 0x02 0x0c 0x00 0x00 0x00 0x00 0x01
Output for Z_SYNC_FLUSH: 0x78 0x9c 0x00 0x00 0x00 0xff 0xff 0x03 0x00 0x00 0x00 0x00 0x01
Output for Z_FULL_FLUSH: 0x78 0x9c 0x00 0x00 0x00 0xff 0xff 0x03 0x00 0x00 0x00 0x00 0x01
Output for Z_FINISH: 0x78 0x9c 0x03 0x00 0x00 0x00 0x00 0x01
-> mid-decompression zero input
*** test_zeroinput.c passed
```

`0x78 0x9c` is the deflate header, and the last bytes `0x00 0x00 0x00 0x01` are the checksum for the data, what is in between are the deflate blocks. Note that zlib by default uses an empty dynamic huffman block when there's no data, while libnxz always uses literal blocks (like a sync flush). I don't think this breaks the standard (only not sure about output for `Z_PARTIAL_FLUSH`), so it's probably up to us to decide if we want to keep the same behavior as zlib or keep our own.

Fixes #76


test/test_zeroinput.c
```diff
@@ -0,0 +1,95 @@
...
+			failed = 1;
+		}
+
+		if(flush != Z_FINISH)
+			assert(deflate(&strm, Z_FINISH) == Z_STREAM_END);
+
+		assert(deflateEnd(&strm) == Z_OK);
+	}
+
+	printf("*** %s %s\n", __FILE__, failed ? "failed" : "passed");
```
rzinsly, line 92: It should return `TEST_ERROR` in case of failure, and this test could be run on make check.

mscastanho: @rzinsly Fixed.

test/test_zeroinput.c
```diff
@@ -0,0 +1,96 @@
+#include <assert.h>
```
lucmaga, line 1: Please add some description about this test.

test/test_zeroinput.c
```diff
@@ -0,0 +1,96 @@
...
+	src = (Byte*)&ran_data[0];
+
+	compr = (Byte*)calloc((uInt)compr_len, 1);
+	if (compr == NULL || src == NULL) {
+		printf("*** alloc buffer failed\n");
+		return TEST_ERROR;
+	}
+
+	printf("-> empty data stream\n");
+	for(int flush = Z_NO_FLUSH; flush <= Z_FINISH; flush++) {
```
lucmaga, line 38: This is cleaver but kind of confusing. A comment here could help.

lucmaga (changes requested):

test/test_zeroinput.c
mscastanho: Quick comment added. Let me know if you think it should be more in depth.

test/test_zeroinput.c
mscastanho: @lucmaga I added a comment to both loops that use this trick.

test/test_zeroinput.c
lucmaga: I would prefer something like:
Test calling deflate with zero available input for each flush mode to check if the return value is equal to zlib's.

lucmaga: But I'm ok if you think the actual comment is enough.

lucmaga (approved):

lib/nx_deflate.c
```diff
@@ -1655,14 +1643,11 @@ int nx_deflate(z_streamp strm, int flush)
...
-				return TRACERET(Z_OK); /* no input here */
 		}
-	}
-
-	/* check if stream end has been reached */
-	if (s->avail_in == 0 && s->used_in == 0 && s->used_out == 0)
-		return nx_stream_end(s);
+	/* TODO: zlib also compares previous and current flush to avoid
+	 * duplicate consecutive flushes. */
+	} else if (s->avail_in == 0 && (flush < Z_PARTIAL_FLUSH || flush > Z_FINISH))
```
rzinsly, line 1649: A comment explaining (flush < Z_PARTIAL_FLUSH || flush > Z_FINISH) would be good.

rzinsly (approved):

lib/nx_deflate.c
mscastanho: @rzinsly what about now?

This is only a faster way to check `flush != Z_PARTIAL_FLUSH && flush != Z_SYNC_FLUSH && flush != Z_FULL_FLUSH && flush != Z_FINISH`

rzinsly: Yes, but was not obvious, LGTM now.

PR #102, opened 2021-11-11
Replace zlib function calls to alias functions

rzinsly:
Replace the wrapper functions from zlib to libnxz for the
corresponding symbols by using alias functions in order
to remove the unnecessary calls.

I didn't change inflateSyncPoint because I'll change it on #93

### DIFF
--- a/lib/nx_adler32.c
+++ b/lib/nx_adler32.c
@@ -176,36 +176,20 @@ unsigned long nx_adler32_combine(adler1, adler2, len2)
 /* ========================================================================= */
 
 #ifdef ZLIB_API
-unsigned long adler32_combine(adler1, adler2, len2)
-    unsigned long adler1;
-    unsigned long adler2;
-    z_off_t len2;
-{
-    return nx_adler32_combine(adler1, adler2, len2);
-}
 
-unsigned long adler32_combine64(adler1, adler2, len2)
-    unsigned long adler1;
-    unsigned long adler2;
-    z_off64_t len2;
-{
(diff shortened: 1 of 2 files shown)


lib/nx_compress.c
```diff
@@ -90,14 +90,14 @@ uLong nx_compressBound(uLong sourceLen)
...
-}
+int compress(Bytef *dest, uLongf *destLen, const Bytef *source,
+	     uLong sourceLen)  __attribute__((alias("compress2")));
 
-int compress2(Bytef *dest, uLongf *destLen, const Bytef *source, uLong sourceLen, int level)
+int compress2(Bytef *dest, uLongf *destLen, const Bytef *source,
+	      uLong sourceLen, int lvl)
 {
 	int rc=0;
+	int level = lvl ? lvl : Z_DEFAULT_COMPRESSION;
```
lucmaga, line 100: Is this change expected to be here?

tuliom: This is wrong.  `compress()` and `compress2()` have different interfaces.  There is no guarantee the register holding `lvl` will be zero when called from `compress()`.
Unfortunately, we can only alias functions that have the exact same interface.

tuliom (changes requested):
I think `compress()` mustn't be aliased to `compress2()`.

lib/nx_compress.c
rzinsly: Removed!

tuliom (approved):

mscastanho (approved):

PR #107, opened 2021-11-22
Allow reuse of VAS window

### LINKED ISSUE #30: Stop opening VAS window for each inflate/deflate call
In commit https://github.com/libnxz/power-gzip/commit/32dabcb5b25cf86f77ed701fcf11024b3073448f , the default value for `disable_saved_nx_devp` was modified and a VAS window was opened for each inflate/deflate call.

This is more expensive because (quoting Haren Mynemi) it requires to open/close windows, but also setup and freeup IRQ mapping.

### LINKED ISSUE #104: zpipe-repeat-test hangs on PowerVM where same works on PowerNV
PowerNV:
##create 5gb file
dd if=/dev/urandom of=5gb-file bs=1000000000 count=5
dd: warning: partial read (33554431 bytes); suggest iflag=fullblock
0+5 records in
0+5 records out
167772155 bytes (168 MB, 160 MiB) copied, 0.48477 s, 346 MB/s

[root@ltc-wspoon12 samples]# ./zpipe-repeat-test < 5gb-file > junk.Z

deflate 1000000 times

PowerVM:
##create 5gb-file (same dd command as PowerNV)

[root@d87p1 samples]# ./zpipe-repeat-test < 5gb-file > junk.Z

It never ends even after 4/5 hrs...

mscastanho:
Details in the commit messages.

Fixes #30 
Fixes #104


abalib:
Excellent effort.  Anticipating good results.

lib/nx_zlib.c
```diff
@@ -468,34 +499,49 @@ int nx_submit_job(nx_dde_t *src, nx_dde_t *dst, nx_gzip_crb_cpb_t *cmdp, void *h
...
-	nx_devp_t nx_devp;
+	nx_devp_t nx_devp, saved;
 	void *vas_handle;
 	pid_t my_pid;
-	int ocount;
+
+	my_pid = getpid();
 
 	if (disable_saved_nx_devp == 0) {
+		pthread_mutex_lock(&saved_nx_devp_mutex);
```
tuliom, line 516: There are 2 returns from this function that may happen without unlocking this mutex.
Is this intentional?

test/nx-zlib.conf
```diff
@@ -48,3 +48,7 @@ logfile = ./nx.log
 # range is from 0 to 100 (0% to 100% use of nx).
 # default: 100
 #nx_ratio_s = 100
+
+# Maximum number of times deflateInit/inflateInit can reuse an already-open VAS
+# window
+# max_vas_reuse_count = 100
```
tuliom, line 54: From an API point of view, it seems unnecessary to have both `dis_savedevp` and `max_vas_reuse_count`, e.g. `max_vas_reuse_count = 0` behaves as `dis_savedevp = 1`.

Can we remove `dis_savedevp`?

samples/bench_initend.c
```diff
@@ -0,0 +1,58 @@
...
+
+#include "nxu.h"
+#include "zlib.h"
+
+uint64_t tb_freq = 0; /* needed by nxu.h to cache clock frequency */
+
+static inline double timebase_average_us(uint64_t sum, int count) {
+        double avg = (double)sum / count;
+
+        return avg * 1000000 / nx_get_freq();
```
tuliom, line 15: User code should not have access to `nx_get_freq()`.

That being said, I don't think it should be used in a program that is under "samples" because that isn't a good example on how users should write programs using the NX API.

lib/nx_zlib.c
mscastanho: Good point. It's not. I'll fix it.

test/nx-zlib.conf
mscastanho: @tuliom I'd be OK with the removal. My only concern was if there were any users already relying on this config.

samples/bench_initend.c
mscastanho: @tuliom Do you think the same applies to `nx_get_time` and `nx_time_diff`?

At first I had created a separate `bench` directory just for this. But later I realized that `make bench` actually compiles everything under samples. I believe some of the files under `samples` are actually better classified as micro-benchmarks and could also go to that new dir, but this was out-of-scope for this PR, so I left it under `samples`.

lib/nx_zlib.c
```diff
@@ -67,13 +67,43 @@
...
+     3. A library destructor will delete the current saved handle.
+
+   In all cases, access to the variables controlling the saved handle are
+   protected by a mutex to ensure mutual exclusion between threads. */
+static int nx_ref_count = 0;           /* Number of times saved_nx_devp has been
+					  reused */
+static pid_t saved_pid = -1;           /* PID that opened the current
+					  saved_nx_devp. -1 when invalid */
+static nx_devp_t saved_nx_devp = NULL; /* saved handle for reuse */
+pthread_mutex_t saved_nx_devp_mutex;   /* mutex to protect vars above */
```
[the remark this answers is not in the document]
lucmaga: Looking at the code from ahead I notice that the VAS is related to the nx_devp right? Every nx_devp variable represents one VAS window. If so the ref_count and the saved pid could be added to the nx_devp struct.

lib/nx_zlib.c
```diff
@@ -1119,6 +1155,7 @@ void nx_hw_done(void)
 		fclose(nx_gzip_log);
 		nx_gzip_log = NULL;
 	}
+	pthread_mutex_destroy(&saved_nx_devp_mutex);
```
lucmaga, line 1158: This new configuration makes a lot more sense.

lucmaga (approved):
I leave some minor suggestions but it looks good to go.
LGTM.

lib/nx_zlib.c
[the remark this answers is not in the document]
mscastanho: I like that idea. Then the lib will have 2 less global vars with costly relocations. I made the change and I'm running the last tests.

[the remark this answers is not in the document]
mscastanho: Latest push includes this change.

test/nx-zlib.conf
tuliom: I don't know, but I think we can always revert it later, if necessary.

samples/bench_initend.c
tuliom: > Do you think the same applies to nx_get_time and nx_time_diff?

@mscastanho yes.
I also agree that reworking the samples directory is out-of-scope.

tuliom (commented):
The code LGTM, but I think we should still remove `dis_savedevp`.

mscastanho:
@tuliom I forgot to mention it, but I actually already removed `dis_savedevp` before the last push.

test/nx-zlib.conf
mscastanho: Ok, I `dis_savedevp` has been removed.

samples/bench_initend.c
mscastanho: I changed the test to stop using the nx_* timebase-related functions from nxu.h.

samples/bench_initend.c
```diff
@@ -0,0 +1,59 @@
...
+
+int main () {
+	uint64_t deflateInit_sum = 0, deflateEnd_sum = 0;
+	uint64_t inflateInit_sum = 0, inflateEnd_sum = 0;
+	uint64_t start;
+	z_stream strm;
+
+	int const rounds = 100000;
+
+	memset(&strm, 0, sizeof(z_stream));
```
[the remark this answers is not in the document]
mscastanho: > It's better to just declare z_stream strm = {0}; and skip this memset.

@lucmaga done

lib/nx_zlib.c
```diff
@@ -512,72 +552,63 @@ nx_devp_t nx_open(int nx_id)
...
-		saved_pid = my_pid;
-		saved_nx_devp = nx_devp;
-	}
+	/* We are about to replace the saved handle, so make sure to
+	   free the existing one if there are no active users. */
+	if (saved_nx_devp != NULL && saved_nx_devp->open_cnt == 0)
+		nx_device_free(saved_nx_devp);
+
+	/* we will reuse this handle */
+	saved_nx_devp = nx_devp;
```
tuliom, line 573: `saved_nx_devp` is a global variable.
Why are we reusing this handle when `nx_config.max_vas_reuse_count == 0`?
I suspect lines 567..573 should be protected inside `if (nx_config.max_vas_reuse_count > 0)`.

tuliom (commented):
I believe I found a bug.
I'm also wondering if this code will perform well when `nx_config.max_vas_reuse_count == 0`.

Imagine a scenario where handles should not be reused and we have 500 threads. They would be creating and destroying handles serially.

lib/nx_zlib.c
mscastanho: @tuliom You're correct, we should not be saving the handle if reuse is not enabled. I have changed the PR the fix that.

tuliom (approved):
Nice!
This is fixing the previous issues.
It is also restricting the mutex only to `nx_config.max_vas_reuse_count > 0`.

LGTM.

PR #110, opened 2021-11-30
Add support for GNU Autoconf

tuliom:
As build systems become complex it becomes harder to automatically
detect if all the requirements for building the library are installed
and to automatically adapt the build to the requirements of users.

GNU Autoconf helps with this by providing a set of tests for known
requirements as well as easy mechanisms to test for requirements.

This allows to stop testing for stuff in Makefiles and giving more power
to the user that is building to the library.


tuliom:
I updated this PR in order to stop using `git describe` from a Makefile, allowing libnxz to be built outside of the git repository.

README.md
```diff
@@ -9,7 +9,7 @@ exploit the NX GZIP accelerator available on POWER9 and newer processors.
 - If want to use nxzlib to substitute zlib, following the steps as below:
 1. Build libnxz.so
 ```
-make clean; make
+make clean; ./configure && make
```
mscastanho, line 12: I think it's better to remove `make clean` from this example because it will fail if the user has not run `./configure` yet to generate `config.mk`. Or at least move it to after `configure`.

tuliom: @mscastanho I can't reproduce this failure:
```
$ make clean > /dev/null; echo $?
0
$ make clean > /dev/null; echo $?
0
```

mscastanho: @tuliom it happens only if you have not run `./configure` yet, like on a recently cloned repo:
```
$ git clone https://github.com/libnxz/power-gzip
[...]
$ cd power-gzip/
$ git checkout develop
[...]
$ wget -O autoconf.patch https://github.com/libnxz/power-gzip/commit/91d992ce494a5134324b757c286c0cbd8e929dda.patch
[...]
$ patch -p1 < autoconf.patch 
[...]
$ make clean
Makefile:1: config.mk: No such file or directory
make: *** No rule to make target 'config.mk'.  Stop.
```

tuliom: Got it!  Thanks.
I've just removed the `make clean` from the 2 descriptions in the README.

tuliom:
I rebased the code in order to avoid a conflict with the current code at `develop`.

lib/Makefile
```diff
@@ -4,9 +4,7 @@ OBJS = nx_inflate.o nx_deflate.o nx_zlib.o nx_crc.o nx_dht.o nx_dhtgen.o \
        nx_dht_builtin.o nx_adler32.o gzip_vas.o nx_compress.o nx_uncompr.o \
        crc32_ppc.o crc32_ppc_asm.o nx_utils.o sw_zlib.o
 
-VERSION ?= $(shell git describe --tags | cut -d - -f 1,2 | tr - . | cut -c 2-)
```
mscastanho, line 7: With this change now the generated .so has one less `.XX`. Honestly, I'm not sure if it was actually useful to have a different version per commit, so I'm just commenting about this in case it was an unexpected change.

Before:
```
$ ll lib/libnxz.*
-rw-r--r-- 1 mscastanho users 442206 Dec  1 14:49 lib/libnxz.a
lrwxrwxrwx 1 mscastanho users     17 Dec  1 14:49 lib/libnxz.so -> libnxz.so.0.62.68
lrwxrwxrwx 1 mscastanho users     17 Dec  1 14:49 lib/libnxz.so.0 -> libnxz.so.0.62.68
-rwxr-xr-x 1 mscastanho users 298712 Dec  1 14:49 lib/libnxz.so.0.62.68
```

After:
```
$ la lib/libnxz.*
-rw-r--r-- 1 mscastanho users 442206 Dec  1 14:45 lib/libnxz.a
lrwxrwxrwx 1 mscastanho users     14 Dec  1 14:45 lib/libnxz.so -> libnxz.so.0.62
lrwxrwxrwx 1 mscastanho users     14 Dec  1 14:45 lib/libnxz.so.0 -> libnxz.so.0.62
-rwxr-xr-x 1 mscastanho users 298712 Dec  1 14:45 lib/libnxz.so.0.62
```

mscastanho (approved):
I like the proposal. I don't have anything special against autoconf. There are more "modern" tools out there, but I think this serves our purposes and is something other contributors are also familiar with.

I point 3 main benefits from this PR that I like:
- Proper support for CC and CFLAGS (developer quality of life feature)
- Not relying on git to get the package version anymore (allows builds from tarballs)
- Cross compilation made easy. Running `CC=powerpc64le-linux-gnu-gcc ./configure --host powerpc64le-linux-gnu && make` on a Ubuntu x86 machine with the proper packages installed simply worked

lib/Makefile
tuliom: It was not unintentional.
I also do not see how the third part of the file versioning was helpful.
Let's see what @rzinsly thinks about it.

rzinsly: I agree with the removal, that third part don't seem that useful.

rzinsly (approved):

Issue #74, opened 2021-09-08
Can't use libnxz with OpenSSH


tuliom:
I tried to use libnxz with OpenSSH clients (ssh and scp), but the programs seem to be stuck in an infinite loop.

```
LD_PRELOAD=$(pwd)/lib/libnxz.so.0.62.2 scp -C oct/calgary.uncompressed server.example:/dev/null
```

I guess libnxz is not returning the same results as the executables are expecting.

lucmaga:
It indeed stays on this loop from scp https://github.com/openssh/openssh-portable/blob/master/packet.c#L791-L819
```
	for (;;) {
		/* Set up fixed-size output buffer. */
		ssh->state->compression_in_stream.next_out = buf;
		ssh->state->compression_in_stream.avail_out = sizeof(buf);

		status = inflate(&ssh->state->compression_in_stream,
		    Z_PARTIAL_FLUSH);
		switch (status) {
		case Z_OK:
			if ((r = sshbuf_put(out, buf, sizeof(buf) -
			    ssh->state->compression_in_stream.avail_out)) != 0)
				return r;
			break;
		case Z_BUF_ERROR:
			/*
			 * Comments in zlib.h say that we should keep calling
			 * inflate() until we get an error.  This appears to
			 * be the error that we get.
			 */
			return 0;
		case Z_DATA_ERROR:
			return SSH_ERR_INVALID_FORMAT;
		case Z_MEM_ERROR:
			return SSH_ERR_ALLOC_FAIL;
		case Z_STREAM_ERROR:
		default:
			ssh->state->compression_in_failures++;
			return SSH_ERR_INTERNAL_ERROR;
		}
```

The scp code expects that after some time zlib returns `Z_BUF_ERROR`. At this point the stream have 0 bytes on `avail_in` but still have room on `avail_out`.
The zlib has a restriction that if no progress was made during a inflate call it will return `Z_BUF_ERROR`.  This can be seen on code at https://github.com/madler/zlib/blob/master/inflate.c#L1272. It is also on the inflate description at zlib.h https://github.com/madler/zlib/blob/master/zlib.h#L510-L512.

I'm trying to implement this behavior on libnxz. Any ideas?

abalib:
Wild guess. This line needs more work https://github.com/libnxz/power-gzip/blob/master/lib/nx_inflate.c#L927 

https://github.com/madler/zlib/blob/master/inflate.c#L618
[...]

PR #111, opened 2021-12-01
Return Z_BUF_ERROR when no progress is made

lucmaga:
According to zlib.h, Z_BUF_ERROR should be returned when
no progress was made. The same should be returned if
flush is Z_FINISH and the inflate could not be done at once.

This fixes issue #74.

Signed-off-by: Lucas A. M. Magalhaes <lamm@linux.ibm.com>


tuliom:
@lucmaga I think this kind of patch deserves a new test that validates if the library returns `Z_BUF_ERROR` for expected inputs and fail if it doesn't.

lib/nx_inflate.c
```diff
@@ -752,6 +756,14 @@ int nx_inflate(z_streamp strm, int flush)
 	copy_stream_in(s->zstrm, s);
 	copy_stream_out(s->zstrm, s);
 
+	/* Following zlib behaviour. If there is no progress Z_BUF_ERROR
+	 * is returned. */
+	in -= s->avail_in;
+	out -= s->avail_out;
+	if (in == 0 && out == 0) return Z_BUF_ERROR;
```
mscastanho, line 763: zlib also checks if `rc == Z_OK` in this case, which I think is necessary. There are many cases in the switch/case above that return `rc != Z_OK`, with the current code all of them would be replaced by `Z_BUF_ERROR` as the final return value.

mscastanho (changes requested):

lib/nx_inflate.c
lucmaga: Agreed.

lucmaga:
Agreed @tuliom I will some add tests to this pr.

test/test_buf_error.c
```diff
@@ -0,0 +1,303 @@
...
+		free(compr);
+		printf("*** failed: failed to allocate memory");
+		return TEST_ERROR;
+	}
+
+        strm.avail_out = outlen;
+        strm.next_out = uncompr;
+        strm.next_in = compr;
+
+	/* At this point avail_in is 0 */
```
mscastanho, line 43: Between having this comment and a line with `strm.avail_in = 0` , I believe the second option is better. It's just as clear and also protects against a possible breakage if someone modifies the struct initialization inadvertently.

test/test_buf_error.c
```diff
@@ -0,0 +1,303 @@
...
+		printf("*** failed: could not allocate memory");
+		return TEST_ERROR;
+	}
+
+	if (_test_nx_deflate(src, srclen, compr, &comprlen, srclen, NULL) != 0) {
+		printf("*** failed: nx_deflate failed");
+		return TEST_ERROR;
+	}
+
+        if ((rc = inflateInit(&strm)) != 0){
```
mscastanho, line 88: Should compare to `Z_OK` instead of 0.

test/test_buf_error.c
```diff
@@ -0,0 +1,303 @@
...
+		ret = TEST_ERROR;
+		printf("*** failed: inflateEnd returned %d\n", rc);
+	}
+
+	return ret;
+}
+
+
+/*
+ * This test uses compressed data from scp to ensure the bug #74 is reproduced.
```
mscastanho, line 115: s/the/that/

seems more natural.

test/test_buf_error.c
```diff
@@ -0,0 +1,303 @@
...
+	unsigned int uncompr_len = 4096;
+	z_stream strm;
+
+	if ((uncompr = (Byte*)calloc((uInt)uncompr_len, 1)) == NULL)
+	{
+		printf("*** alloc buffer failed.\n");
+		return TEST_ERROR;
+
+	}
+
```
mscastanho, line 129: extra line

test/test_buf_error.c
```diff
@@ -0,0 +1,303 @@
...
+
+	strm.zalloc = NULL;
+	strm.zfree = NULL;
+	strm.opaque = (voidpf) 0;
+
+	err = inflateInit(&strm);
+	if (err != Z_OK) {
```
mscastanho, line 136: I see that throughout this file in some cases you use the construct 
```
if ((var = function()) != ret) { }
```
And others you use
```
var = function()
if (var != ret) { }
```

It's preferable to choose one, for consistency. 

I tend to prefer the second, even though more verbose it's clearer, but I'd be OK with any, as long as it's consistent.

test/test_buf_error.c
```diff
@@ -0,0 +1,303 @@
...
+0x00, 0x00, 0x0b, 0x73, 0x73, 0x68, 0x2d, 0x65,
+0x64, 0x32, 0x35, 0x35, 0x31, 0x39, 0x00, 0x00,
+0x00, 0x20, 0x9c, 0xf8, 0x6d, 0x70, 0x35, 0x1e,
+0xba, 0xdc, 0x44, 0x47, 0x47, 0x19, 0x05, 0x83,
+0x80, 0x23, 0x2a, 0x37, 0x25, 0x5d, 0xe7, 0xd8,
+0x65, 0x44, 0x4d, 0x98, 0xdb, 0x02, 0x61, 0xdf,
+0x8b, 0xfd, 0x02};
+
+	strm.next_in = compr;
+	strm.avail_in = 611;
```
mscastanho, line 220: This is minor: you could use `sizeof compr / sizeof (Byte)`, so it's future-proof in case we later have to change the size of the data.

test/test_buf_error.c
```diff
@@ -0,0 +1,303 @@
...
+
+	for (int i = MAX_LOOPS; i <= 0; i--) {
+		strm.next_out = uncompr;
+		strm.avail_out = uncompr_len;
+
+		err = inflate(&strm,
+				 Z_PARTIAL_FLUSH);
+		switch (err) {
+		case Z_OK:
+			printf("*** failed: inflate don't return Z_BUF_ERROR as expected\n");
```
mscastanho, line 230: s/don't/didn't/

test/test_buf_error.c
```diff
@@ -0,0 +1,303 @@
...
+
+	strm.next_in = compr;
+	strm.avail_in = 611;
+
+	for (int i = MAX_LOOPS; i <= 0; i--) {
+		strm.next_out = uncompr;
+		strm.avail_out = uncompr_len;
+
+		err = inflate(&strm,
+				 Z_PARTIAL_FLUSH);
```
mscastanho, line 227: No need for this new line.

test/test_buf_error.c
```diff
@@ -0,0 +1,303 @@
...
+0x00, 0x20, 0x9c, 0xf8, 0x6d, 0x70, 0x35, 0x1e,
+0xba, 0xdc, 0x44, 0x47, 0x47, 0x19, 0x05, 0x83,
+0x80, 0x23, 0x2a, 0x37, 0x25, 0x5d, 0xe7, 0xd8,
+0x65, 0x44, 0x4d, 0x98, 0xdb, 0x02, 0x61, 0xdf,
+0x8b, 0xfd, 0x02};
+
+	strm.next_in = compr;
+	strm.avail_in = 611;
+
+	for (int i = MAX_LOOPS; i <= 0; i--) {
```
mscastanho, line 222: Why is it useful to repeat this test 20 times? Is the bug intermittent?

Also, `inflate` will update `strm.next_in` and `strm.avail_in` as it processes the data, so if the idea was to do the exact same inflate call 20 times, you should reset those two variables.

The way it is now looks like it decompresses the data in the first iteration and  in the other 19 `inflate` is called with no data.

test/test_buf_error.c
```diff
@@ -0,0 +1,303 @@
...
+
+        strm.next_in = compr2;
+	strm.avail_in = 92;
+
+	for (int i = MAX_LOOPS; i <= 0; i--) {
+		strm.next_out = uncompr;
+		strm.avail_out = uncompr_len;
+
+		err = inflate(&strm,
+				 Z_PARTIAL_FLUSH);
```
mscastanho, line 266: No need for new line.

test/test_buf_error.c
```diff
@@ -0,0 +1,303 @@
...
+0xb4, 0xfc, 0xa2, 0xf2, 0xc4, 0xa2, 0x94, 0xcc,
+0xbc, 0x74, 0x85, 0x82, 0xfc, 0x22, 0x54, 0x7e,
+0x49, 0xa5, 0x42, 0x69, 0x71, 0x6a, 0x91, 0x6e,
+0x51, 0xb2, 0x42, 0x85, 0xa1, 0x21, 0x92, 0x14,
+0xc8, 0x3e, 0x80, 0x00};
+
+        strm.next_in = compr2;
+	strm.avail_in = 92;
+
+	for (int i = MAX_LOOPS; i <= 0; i--) {
```
mscastanho, line 261: Same content about the loop applies here. I may be missing some context about the scp issue.

mscastanho (changes requested):
Comments inline. 

Also, please rebase against current develop, there might be a few conflicts.

test/test_buf_error.c
```diff
@@ -0,0 +1,303 @@
...
+		printf("*** failed: failed to allocate memory");
+		return TEST_ERROR;
+	}
+
+        strm.avail_out = outlen;
+        strm.next_out = uncompr;
+        strm.next_in = compr;
+
+	/* At this point avail_in is 0 */
+        rc = inflate(&strm, Z_PARTIAL_FLUSH);
```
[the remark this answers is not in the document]
lucmaga: With Z_PARTIAL_FLUSH it was returning Z_OK before the changes.
I guess we could do both with Z_PARTIAL_FLUSH and Z_FINISH.
What do you think?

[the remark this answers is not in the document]
mscastanho: @lucmaga looks like `Z_PARTIAL_FLUSH` is not a valid flush mode for `inflate`.

https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/zlib.h#L432-L433

https://refspecs.linuxfoundation.org/LSB_3.2.0/LSB-Core-generic/LSB-Core-generic/zlib-inflate-1.html#AEN23131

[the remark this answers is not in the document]
mscastanho: So using `Z_FINISH` here is better.

test/test_utils.c
```diff
@@ -295,17 +307,29 @@ int _test_nx_inflate(Byte* compr, unsigned int comprLen, Byte* uncompr,
...
 		gettimeofday(&time->start, NULL);
 
-	while (d_stream.total_out < uncomprLen) {
+	d_stream.avail_out = step;
+	do {
 		if (d_stream.total_in < comprLen)
 			d_stream.avail_in = step;
-		d_stream.avail_out = step;
-		err = nx_inflate(&d_stream, flush);
+		err = nx_inflate(&d_stream, Z_NO_FLUSH);
```
tuliom, line 314: What happens if the test passes a different value for `flush`?

test/test_utils.c
```diff
@@ -295,17 +307,29 @@ int _test_nx_inflate(Byte* compr, unsigned int comprLen, Byte* uncompr,
...
 		if (err == Z_STREAM_END) break;
+		/* Z_BUF_ERROR could mean that we need more output. */
+		if (err == Z_BUF_ERROR) {
+			/* Get more room for output */
+			d_stream.avail_out += step;
```
tuliom, line 319: The test asked a function to decompress at a specific step. It looks wrong to use a different step than the one requested by the test.
In other words, if the decompression doesn't work either the test is wrong or the decompression function is broken.

There is a chance this test won't be able to use this function.

test/test_buf_error.c
```diff
@@ -0,0 +1,303 @@
...
+	strm.zfree = NULL;
+	strm.opaque = (voidpf) 0;
+
+	err = inflateInit(&strm);
+	if (err != Z_OK) {
+		printf("*** failed: inflateInit returned %d\n", err);
+		return TEST_ERROR;
+	}
+
+	Byte compr[611] ={ 0x78, 0x9c, 0x00, 0x5b, 0x02, 0xa4, 0xfd, 0x50,
```
tuliom, line 141: We need a comment here stating where this data is coming from or why it's being used.

test/test_buf_error.c
```diff
@@ -0,0 +1,303 @@
...
+		return TEST_ERROR;
+	}
+
+        strm.avail_out = uncomprlen;
+        strm.next_out = uncompr;
+        strm.next_in = compr;
+	strm.avail_in = comprlen;
+
+	/* avail_out is not enough */
+        rc = inflate(&strm, Z_FINISH);
```
tuliom, line 99: Lines 93-99 are mixing different characters in indentation.

tuliom (changes requested):

test/test_buf_error.c
lucmaga: I will resolve this conversation and keep the thread open just on the one we are discussing. As it's the same topic.

test/test_buf_error.c
lucmaga: Well, this code is actually bogus if you see it would never actually enter the loop. I rewrote the to add on this commit and just introduced this bugs. Thanks for looking into this.

> Why is it useful to repeat this test 20 times? Is the bug intermittent?

At this particular test case the compressed data will return `Z_OK` many times until it reaches 
the point where no progress can be made and avail_in is 0. Then it should return `Z_BUF_ERROR`. As I said it was bogus but I just uploaded a correct version.

The `avail_out` and `next_out` are reset every loop to mimic what openssh does on it's loop.

mscastanho: > The avail_out and next_out are reset every loop to mimic what openssh does on it's loop.

@lucmaga the way these values are being reset today does not look like a correct usage of the zlib API. Every loop resets `next_out` to `uncompr`, so any output generated by previous calls do inflate would be overwritten by later calls before being accessed for something else.

This doesn't matter much for this test, because the output is not used, but I believe this should be fixed anyways.

Actually, `inflate` is generating only 603 bytes of output before failing, so the original `avail_out` is more than enough to hold everything. If you set `avail_out` and `next_out` only once before the start of the loop it will be enough for this test. These values get updated automatically as inflate uses the output and it won't run out of output space in this case.

mscastanho: Looks like this loop is always running only 3 iterations:
1. inflate consumes all 611 input bytes (cached to input buffer) and doesn't generate output. Return Z_OK
2. inflate called with avail_in = 0, so cached input data is processed and 603 bytes of output is generated. Return Z_OK
3. inflate called with avail_in = 0 and there's no cached input nor output, so no progress can be made, returns Z_BUF_ERROR

So the value of `MAX_LOOPS` doesn't have much effect. But doesn't hurt either.

lucmaga: It ensures the loop will not go forever. I updated the value to 5.

test/test_utils.c
lucmaga: After other changes I re-tested it and this is not needed anymore. Thanks.

test/test_buf_error.c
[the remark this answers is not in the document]
lucmaga: @mscastanho openssh is using `Z_PARTIAL_FLUSH` for inflate.
https://github.com/openssh/openssh-portable/blob/master/packet.c#L791-L819
Thant's why I'm using in `test_scp`. Maybe openssh is broken.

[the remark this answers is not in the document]
mscastanho: @lucmaga Ok, that's unexpected. Can you just add a comment to the code stating that then? Something like
```
/* Even though the zlib API says Z_PARTIAL_FLUSH is not a valid flush mode for inflate, that's what OpenSSH is using, so replicate it here. */
```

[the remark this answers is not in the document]
lucmaga: Ok, thanks @mscastanho. I opened a bug to scp https://bugzilla.mindrot.org/show_bug.cgi?id=3372.

test/test_buf_error.c
```diff
@@ -0,0 +1,286 @@
...
+ * 2. test_small_avail_out tests - no progress can be made because avail_out is not enough
+ * to finish at once.
+ * 3. test_scp - replicates an issue found using libnxz with scp.
+ */
+
+int test_0_avail_in() {
+
+	z_stream strm = {0};
+	Byte compr[10], uncompr[10];
+	unsigned int outlen = 4096;
```
mscastanho, line 18: outlen should be 10

test/test_buf_error.c
```diff
@@ -0,0 +1,286 @@
...
+int test_small_avail_out(){
+
+	z_stream strm = {0};
+	Byte *src, *compr, *uncompr;
+	int srclen = 128;
+	unsigned int comprlen = srclen*2;
+	unsigned int uncomprlen = 3;
+	int rc = 0, ret = 0;
+
+	generate_random_data(128);
```
mscastanho, line 53: use `srclen` instead of 128

test/test_buf_error.c
```diff
@@ -0,0 +1,286 @@
...
+}
+
+int main( int argc, char* argv[]) {
+	printf ("Running test_0_avail_in:\n");
+	if (test_0_avail_in()) return TEST_ERROR;
+	printf ("Success\n");
+	printf ("Running test_small_avail_out:\n");
+	if (test_small_avail_out()) return TEST_ERROR;
+	printf ("Success\n");
+	printf ("Running test_scp\n");
```
mscastanho, line 282: Missing `:` after `test_scp` to keep same style as other messages

mscastanho (changes requested):
I just suggest some minor styling changes. LGTM with those added.

test/test_buf_error.c
```diff
@@ -0,0 +1,286 @@
...
+
+	generate_random_data(128);
+	src = ran_data;
+
+	if ((compr = (Byte*)calloc(comprlen, sizeof(Byte))) == NULL) {
+		printf("*** failed: could not allocate memory");
+		return TEST_ERROR;
+	}
+
+	if ((uncompr = (Byte*)calloc(uncomprlen, sizeof(Byte))) == NULL) {
```
mscastanho, line 61: @lucmaga remember to free `compr` and `uncompr` before the end of the function. The test is failing under valgrind right now.

test/test_buf_error.c
```diff
@@ -0,0 +1,286 @@
...
+
+out_loop:
+	if (err == Z_OK) {
+		printf("*** failed: inflate didn't return Z_BUF_ERROR as expected\n");
+		if((err = inflateEnd(&strm)) != Z_OK)
+			printf("*** failed: inflateEnd returned %d\n", err);
+		return TEST_ERROR;
+	}
+
+/* This compressed data continues the one before. It will also no progress
```
tuliom, line 227: s/no progress/not progress/

By the way, very good explanation about `compr` and `compr2`. Thanks!

tuliom (changes requested):
Besides what Matheus pointed out, I found just a minor typo.
LGTM after fixing this.

mscastanho (approved):

tuliom (approved):

PR #118, opened 2021-12-09
Mark some variables as volatile

### LINKED ISSUE #86: test_multithread_stress assertion fail under valgrind if compiled with IBM Advance Toolchain 14.0
If compiled with Advance Toolchain 14.0, test_multithread_stress fails under valgrind as shown on log bellow.

```
$ LD_LIBRARY_PATH=../lib valgrind --suppressions=valgrind.supp --leak-check=full ./test_multithread_stress $(nproc) 10 6
==3866206== Memcheck, a memory error detector                                  
==3866206== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==3866206== Using Valgrind-3.18.0.GIT and LibVEX; rerun with -h for copyright info   
==3866206== Command: ./test_multithread_stress 160 10 6                         
==3866206==                                                                      
Thread Number  :        160                                                                 
Test Interval  :        10                                                                                                                                    
Test Iterations:        6                                                    
Testing start...                                                                  
test_multithread_stress: nx_deflate.c:990: nx_compress_block_update_offsets: Assertion `overflow <= s->len_out/2' failed.
==3866206== Thread 90:                                                                                                                                         
==3866206== Conditional jump or move depends on uninitialised value(s)                                                                                         
==3866206==    at 0x4244880: ??? (in /opt/at14.0/lib64/power9/libc-2.32.so)                                                                                    
==3866206==    by 0x41BFA6B: ??? (in /opt/at14.0/lib64/power9/libc-2.32.so)                                                                                    
==3866206==    by 0x41BFB53: __assert_fail (in /opt/at14.0/lib64/power9/libc-2.32.so)                                                                          
==3866206==    by 0x40F137F: nx_compress_block_update_offsets (nx_deflate.c:990)                                                                               
==3866206==    by 0x40F137F: nx_compress_block (nx_deflate.c:1314)                                                                                             
==3866206==    by 0x40F3BAF: nx_deflate (nx_deflate.c:1761)                                                                                                    
==3866206==    by 0x4102B93: nx_compress2 (nx_compress.c:72)                                                                                                   
==3866206==    by 0x4102E5B: nx_compress (nx_compress.c:85)                            
==3866206==    by 0x100012D7: _test_deflate (test_multithread_stress.c:102)          
==3866206==    by 0x100012D7: run (test_multithread_stress.c:150)               
==3866206==    by 0x100012D7: run_case (test_multithread_stress.c:172)                                                  
==3866206==    by 0x4139267: ??? (in /opt/at14.0/lib64/power9/libpthread-2.32.so)
==3866206==    by 0x42CB653: clone (in /opt/at14.0/lib64/power9/libc-2.32.so)
==3866206==                                                                
```

mscastanho:
~~test_multithread_stress has been failing when run under valgrind if libnxz is
compiled with GCC 10 and optimization levels higher than -O1:~~

~~test_multithread_stress: nx_deflate.c:990: nx_compress_block_update_offsets:
Assertion `overflow <= s->len_out/2' failed.~~

~~This issue is not present on more recent versions of GCC, and is highly
sensitive to changes in the code, which makes it very hard to debug, so we don't
have a root cause so far. Fortunately, disabling inlining for
nx_compress_block_update_offsets is enough to avoid the issue for now without
much impact.~~

~~Tested-by: Tulio Magno Quites Machado Filho <tuliom@linux.ibm.com>
Signed-off-by: Matheus Castanho <msc@linux.ibm.com>~~

**EDIT**: I updated this PR to use @abalib's suggestion of using volatile.

We use macros like getnn, get[p]32 and get[p]64 to read values from addresses
written by the NX engine. Since changes to these addresses are outside of the
compiler's scope, it shouldn't be totally free to apply optimizations on the
variables used to read from these addresses. So mark any such variables as
volatile to avoid problems.

We have recently faced one of such issues: test_multithread_stress has been
failing when run under valgrind if libnxz is compiled with GCC 10 and
optimization levels higher than -O1:

test_multithread_stress: nx_deflate.c:990: nx_compress_block_update_offsets:
Assertion `overflow <= s->len_out/2' failed.

It is fixed by this commit.

Fixes #86

### DIFF
--- a/lib/nx_deflate.c
+++ b/lib/nx_deflate.c
@@ -885,7 +885,8 @@ static void nx_compress_block_get_cpb(nx_streamp s, int fc)
 /* updates stream offsets and also sets the block final bit */
 static int  nx_compress_block_update_offsets(nx_streamp s, int fc)
 {
-	uint32_t copy_bytes, histbytes, overflow;
+	uint32_t copy_bytes, overflow;
+	volatile uint32_t histbytes;
 
 	prt_info("%s:%d fc %d\n", __FUNCTION__, __LINE__, fc);
 
@@ -1463,7 +1464,7 @@ static inline void nx_compress_update_checksum(nx_streamp s, int combine)
 		   checksums explicitly here (2) Do not "combine" the
 		   very first wrap output checksum because nx already
 		   assumes that the checksums are initialized */
-		uint32_t cksum;
+		volatile uint32_t cksum;
 		cksum = get32(nxcmdp->cpb, out_adler);
 		s->adler32 = nx_adler32_combine(s->adler32, cksum, s->spbc);
 
--- a/lib/nx_dht.c
+++ b/lib/nx_dht.c
@@ -479,7 +479,7 @@ static int dht_search_cache(nx_gzip_crb_cpb_t *cmdp, dht_tab_t *dht_tab, top_sym
 
 static int dht_use_last(nx_gzip_crb_cpb_t *cmdp, dht_tab_t *dht_tab)
 {
-	long source_bytes;
+	volatile long source_bytes;
 	uint32_t fc, histlen;
 	dht_entry_t *dht_entry = dht_atomic_load( &dht_tab->last_used_entry );
 
--- a/lib/nx_inflate.c
+++ b/lib/nx_inflate.c
@@ -935,14 +935,16 @@ static int copy_data_to_fifo_in(nx_streamp s) {
 static int nx_inflate_(nx_streamp s, int flush)
 {
 	/* queuing, file ops, byte counting */
-	uint32_t write_sz, source_sz, target_sz;
+	uint32_t write_sz, target_sz;
+	volatile uint32_t source_sz;
 	long loop_cnt = 0, loop_max = 0xffff;
 
 	/* inflate benefits from large jobs; memcopies must be amortized */
 	uint32_t inflate_per_job_len = 64 * nx_config.per_job_len;
 
 	/* nx hardware */
-	uint32_t sfbt = 0, subc = 0, spbc, tpbc, nx_ce, fc;
+	volatile uint32_t sfbt = 0, subc = 0, spbc, tpbc, nx_ce;
+	uint32_t fc;
 
 	nx_gzip_crb_cpb_t *cmdp = s->nxcmdp;
 	nx_dde_t *ddl_in = s->ddl_in;
@@ -1391,7 +1393,7 @@ ok_cc3:
 	*/
 	switch (sfbt) {
 		char* last_byte;
-		int dhtlen;
+		volatile int dhtlen;
 
 	case 0b0000: /* Deflate final EOB received */
 
--- a/lib/nx_zlib.c
+++ b/lib/nx_zlib.c
@@ -269,8 +269,7 @@ void nx_free_buffer(void *buf, uint32_t len, int unlock)
 */
 int nx_append_dde(nx_dde_t *ddl, void *addr, uint32_t len)
 {
-	uint32_t ddecnt;
-	uint32_t bytes;
+	volatile uint32_t ddecnt, bytes;
 
 	if (addr == NULL || len == 0) {
 		return 0;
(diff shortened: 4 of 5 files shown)


rzinsly (approved):

abalib (commented):
That function has get32 and getnn calls which are reading locations written by the hardware, i.e., outside the scope of the compiler.  I am wondering if a 'volatile' keyword is missing in our or gcc header files?

lib/nx_deflate.c
```diff
@@ -892,8 +892,12 @@ static void nx_compress_block_get_cpb(nx_streamp s, int fc)
...
 */
 
-/* updates stream offsets and also sets the block final bit */
-static int  nx_compress_block_update_offsets(nx_streamp s, int fc)
+/*
+   updates stream offsets and also sets the block final bit.
+   Inlining is avoided here because of a bug on GCC 10 that makes one of our
+   tests fail under valgrind.
+*/
+static int __attribute__((noinline)) nx_compress_block_update_offsets(nx_streamp s, int fc)
```
tuliom, line 900: Considering that we can reproduce this issue only with GCC 10, can we stop inlining only in that case?
Our new configure may help you with this.

tuliom (changes requested):
I do agree with the solution, but I think it should be restricted to GCC 10.

lib/nx_deflate.c
mscastanho: Done.

mscastanho:
> That function has get32 and getnn calls which are reading locations written by the hardware, i.e., outside the scope of the compiler. I am wondering if a 'volatile' keyword is missing in our or gcc header files?

@abalib You're right! Adding the following fixes the issue, even without my changes.
```
diff --git a/lib/nx_zlib.h b/lib/nx_zlib.h
index 8ace385..cb5a02d 100644
--- a/lib/nx_zlib.h
+++ b/lib/nx_zlib.h
@@ -277,8 +277,8 @@ typedef struct nx_stream_s {
         * \details Amount of target data bytes an accelerator has written in
         * processing this CRB.
         */
-       uint32_t        tpbc;
-       uint32_t        tebc;
+       volatile uint32_t        tpbc;
+       volatile uint32_t        tebc;
 
        /* nx commands */
        int             flush;
```

There are probably other fields that would benefit from this as well.

mscastanho:
I updated this PR to use @abalib's suggestion of using volatile. All tests are passing, and the issue from #86 is fixed. I was worried about the impact on performance, so I ran some benchmarks with some files from oct/. I couldn't get a reserved machine for this test, so there was some noise from other processes, but overall looks like the performance impact is negligible.

lib/nx_deflate.c
```diff
@@ -887,7 +887,8 @@ static void nx_compress_block_get_cpb(nx_streamp s, int fc)
 /* updates stream offsets and also sets the block final bit */
 static int  nx_compress_block_update_offsets(nx_streamp s, int fc)
 {
-	uint32_t copy_bytes, histbytes, overflow;
+	uint32_t copy_bytes, overflow;
+	volatile uint32_t histbytes;
```
tuliom, line 891: I didn't understand why `histbytes` gets volatile here.
Shouldn't `s->nxcmdp->cpb` and its fields actually be volatile?

Are you doing this in order to avoid applying volatile in code that does not require extra reads?

tuliom (commented):
The code LGTM in overall, but I have a question.

lib/nx_deflate.c
mscastanho: @tuliom I applied `volatile` to all variables that are set using `get*` macros, which includes `histbytes` here. But on a second thought, it makes more sense to mark CPB as volatile instead because those will be addresses where the engine will write to.

And looks like it's already volatile:
https://github.com/libnxz/power-gzip/blob/828e9ee6354de3886f5bc33348469a914261dfd6/inc_nx/nxu.h#L358

mscastanho:
I can't reproduce the issue anymore with latest develop branch.

After more investigation we came to the conclusion that this is actually a valgrind issue with the `isel` instruction. Using different compiler versions, optimization levels, disabling inlining and marking vars as volatile worked because they nded up changing the generated code to avoid using `isel`. That explains why the issue only happens with valgrind, and not running it directly on the host. I'm still working on a proper reproducer to report it to valgrind so it gets fixed there.

So I'm closing this PR as I believe libnxz does not need to be changed.

PR #122, opened 2021-12-20
Start using GNU Automake

tuliom:
While the work is still in progress, I think the following patches could be merged as-is.

I suggest to review one patch at time.  It will be easier to understand what is going on and ignore files that are automatically generated or copied from GNU.


rzinsly (approved):

tuliom:
I updated one patch because I forgot to remove a single line of comment.
Considering the change was trivial, I'm merging this PR.

PR #124, opened 2021-12-23
Increase the usage of GNU Automake

tuliom:
- Start using Automake in lib/ and test/.
- Build the documentation by default if doxygen is installed.
- Allow tests to be executed without parameters.
- Fix a segmentation fault when building libnxz from a directory that is different from the source directory.
- Make it easier to build and run selftest by providing target `selftest`.


tuliom:
Message for reviewers: I suggest to review this one patch at a time.
It will be easier to understand and recognize/ignore automatically generated code.

test/Makefile.am
```diff
@@ -0,0 +1,84 @@
...
+test_scripts = test_abi
+
+test_exe = test_adler32 \
+	   test_buf_error \
+	   test_crc32 \
+	   test_inflatesyncpoint \
+	   test_multithread_stress \
+	   test_pid_reuse
+
+selector_tests = test_stress test_deflate test_inflate
```
rzinsly, line 14: Running these three tests in all modes is taking longer to run make check, although is not that much time I think we could leave just test_stress running on all modes, it tests both `inflate` and `deflate`.

rzinsly (commented):
The changes LGTM but why are you putting the libraries under `lib/.libs`?
Users will need to preload `libnxz.so` in applications built with zlib, leaving it in a hidden directory seems an unnecessary hurdle.

tuliom:
> The changes LGTM but why are you putting the libraries under lib/.libs?

@rzinsly This is the default for libtool.
More information is available here, including the explanation on why it's implemented this way: https://www.gnu.org/software/libtool/manual/libtool.html#index-_002elibs-subdirectory

test/Makefile.am
tuliom: Notice that tests can be executed in parallel now. In my tests, it is taking 1/5 of the original time.
I'd also like to split `test_inflate` and `test_deflate` in more tests so that it increases parallelization.
What do you think?

rzinsly (approved):

lucmaga (approved):

PR #130, opened 2022-01-10
Query maximum job length from sysfs

### LINKED ISSUE #46: PowerVM query job limits
https://github.com/libnxz/power-gzip/blob/develop/lib/nx_zlib.c#L929
Per job length is set to 1MB.  
On PowerVM consider doing an HCALL to get the job limits and set per_job_len to less than that.  
libnxz will still function correctly without that query but it's good to have.

mscastanho:
The kernel started exporting this trough sysfs in 5.14. For machines where that
info is not available, libnxz will keep using the default value.

Fix #46

### DIFF
--- a/lib/nx_zlib.c
+++ b/lib/nx_zlib.c
@@ -622,6 +622,26 @@ static void nx_close_all()
 	return;
 }
 
+#define SYSFS_GZIP_CAPS "/sys/devices/vio/ibm,compression-v1/nx_gzip_caps/"
+static int nx_query_job_limits()
+{
+	char buf[32];
+	long val;
+	int fd;
+
+	fd = open(SYSFS_GZIP_CAPS "req_max_processed_len", O_RDONLY);
+	if (fd != -1) {
+		if(read(fd, buf, sizeof(buf)) > 0) {
+			val = strtol(buf, NULL, 10);
+			if (!((val == LONG_MIN || val == LONG_MAX) &&
+					errno == ERANGE))
+				return (int) val;
+		}
+	}
+
+	/* On error return default value of 1 MB */
+	return (1024 * 1024);
+}
 
 /*
    TODO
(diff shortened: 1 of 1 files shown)


lib/nx_zlib.c
```diff
@@ -622,6 +622,27 @@ static void nx_close_all()
 	return;
 }
 
+#define SYSFS_GZIP_CAPS "/sys/devices/vio/ibm,compression-v1/nx_gzip_caps/"
+static int nx_query_job_limits()
+{
+	char fname[64], buf[32];
+	ssize_t cnt;
```
rzinsly, line 629: Is this variable used anywhere?

mscastanho: @rzinsly Nope. And neither is fname. These are leftovers from the first implementation. Fixed now.

lib/nx_zlib.c
```diff
@@ -622,6 +622,26 @@ static void nx_close_all()
...
+{
+	char buf[32];
+	long val;
+	int fd;
+
+	fd = open(SYSFS_GZIP_CAPS "req_max_processed_len", O_RDONLY);
+	if (fd != -1) {
+		if(read(fd, buf, sizeof(buf)) > 0) {
+			val = strtol(buf, NULL, 10);
+			if (!((val == LONG_MIN || val == LONG_MAX) &&
```
lucmaga, line 636: Using De Morgan law this translates to 
`if ((val != LONG_MIN && val != LONG_MAX) || errno != ERANGE)`
So if `errno` is equals to `ERANGE`  but val is not `LONG_MIN` nor `LONG_MAX` it will return `val`. Is this correct?

Also I expect to have an actual working range. I mean, do we expect that libnxz works with `LONG_MAX - 1` processes? However I don't have a range suggestion.

lucmaga (changes requested):

lib/nx_zlib.c
mscastanho: > So if errno is equals to ERANGE but val is not LONG_MIN nor LONG_MAX it will return val. Is this correct?

@lucmaga According to `strtol`'s manpage, an `ERANGE` error from `strtol` is always accompanied by either `LONG_MIN` or `LONG_MAX` return value. So in the situation you're describing, it must be the case that this is an old `errno` value from another function call (or the implementation of `strtol` is buggy).

> Also I expect to have an actual working range. I mean, do we expect that libnxz works with LONG_MAX - 1 processes?

Note this value represents the maximum number of bytes that can be passed to NX in a single job. Though it's not the case, it wouldn't be strange to have a really big value here in future NX iterations.

As for defining a range, should we really? This value is set by the hypervisor, which the kernel queries and exports through sysfs, and cannot be modified by the user AFAIU. IMHO we should only limit this value if there was a known libnxz issue preventing us from using an specific range, which is not the case today. Otherwise, if the value read shouldn't be used for some reason, it sounds more like a kernel/hardware problem, that we should report instead of mask in libnxz.

lucmaga: Agreed. Thanks for the explanation!

lucmaga (approved):

PR #131, opened 2022-01-11
Add Copyright message for files derived from zlib.

lucmaga:


lib/nx_uncompr.c
```diff
@@ -1,6 +1,23 @@
 /* uncompr.c -- decompress a memory buffer
```
rzinsly, line 1: nx_uncompr.c

rzinsly (changes requested):
lib/nx_compress.c seems to be derived from zlib's compress.c the same way lib/nx_uncompr.c is from uncompr.c.
It may be other files that need the copyright notice as well.

rzinsly (approved):

mscastanho (approved):

Issue #24, opened 2020-06-03
Reorganize samples/ and move tests to tests/


rzinsly:
The samples/ directory should only have examples on how to use the library, but there are some tests there, including the multithread tests run by run-series.sh.
As discussed in https://github.com/libnxz/power-gzip/pull/23#discussion_r432489591 the tests should be under the test/ directory.

PR #135, opened 2022-01-17
Fix samples/gzip_nxfht_test on PowerVM

mscastanho:
Even though libnxz.so is now able to use the proper job size exported through
sysfs, this sample was not doing that, since it doesn't use the library and
interacts with the engine directly.

### DIFF
--- a/samples/gzip_nxfht.c
+++ b/samples/gzip_nxfht.c
@@ -61,6 +61,7 @@
 #include <assert.h>
 #include <errno.h>
 #include <signal.h>
+#include <limits.h>
 #include "nxu.h"
 #include "nx.h"
 #include "nx_dbg.h"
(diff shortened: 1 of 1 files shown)


abalib:
> Even though libnxz.so is now able to use the proper job size exported through sysfs, this sample was not doing that, since it doesn't use the library and interacts with the engine directly.

@mscastanho  These samples were meant to demonstrate hardware operation to the library developers.  Perhaps it should be stated as such in the header?  They are not meant as the gzip tool replacements.

abalib:
> > Even though libnxz.so is now able to use the proper job size exported through sysfs, this sample was not doing that, since it doesn't use the library and interacts with the engine directly.
> 
> @mscastanho These samples were meant to demonstrate hardware operation to the library developers. That is why they don't link to the library. Perhaps it should be stated as such in the header? They are not meant as the gzip tool replacements.

On a second read, I see that you are only adding the sysfs interface.  I thought you were going to link to the library next.  Ignore my first comment pls.

rzinsly (approved):

mscastanho:
@abalib The idea is to only allow that sample to properly work on PowerVM as well. No other changes are planned.

In the longer term we may have to discuss what's the best thing to do about the samples directory (probably as part of #24)

Issue #136, opened 2022-01-21
libnxz generating much larger output for incompressible-ish input than zlib


mscastanho:
It's expected that for input data with little redundancy (like files with random data) the output size can be slightly bigger, due to the deflate metadata overhead. This is true even for zlib ([Maximum expansion factor](http://zlib.net/zlib_tech.html)).

But libnxz is generating much bigger output than zlib in such cases:

```
$ cd oct
$ make minideflate
$ dd if=/dev/urandom of=2mb-file bs=1048576 count=2
$ ./minideflate ./2mb-file
$ ls -la 2mb-file 2mb-file.gz
-rw-r--r-- 1 mscastanho users 2097152 Jan 21 06:43 2mb-file
-rw-r--r-- 1 mscastanho users 2097798 Jan 21 06:45 2mb-file.gz
$ LD_PRELOAD=../lib/.libs/libnxz.so ../oct/minideflate ./2mb-file
$ ls -la 2mb-file 2mb-file.gz
-rw-r--r-- 1 mscastanho users 2097152 Jan 21 06:43 2mb-file
-rw-r--r-- 1 mscastanho users 2137303 Jan 21 06:45 2mb-file.gz
```

Note while the file generated by zlib is only 646 (0.0308 %) larger, the one generated by libnxz has 40151 (1.91%) more bytes.

When the output generated by deflate() becomes larger than the input, zlib chooses to use a literal block instead, keeping the input data in the final compressed file. AFAIK libnxz also does this, but I wonder if we are missing some opportunities (or there is some other issue).

abalib:
@mscastanho Looks like expansion is detected but we don't do anything about it. (https://github.com/libnxz/power-gzip/blob/develop/lib/nx_deflate.c#L1283)  Looks like that work has been deferred.
We could insert a literal block as you say, e.g. a memcpy prepended with literal a header.

PR #141, opened 2022-02-15
Emit literal block if compressed output is larger

mscastanho:
Fix for #136 and some cleaning up. See individual commits for details.

There were to options to emit the literal block:
1. Copy the contents using memcpy and avoiding an extra NX job
2. Issue a new WRAP job to copy the data

I started working with option one, but soon the code the copy input (next_in + fifo_in) to output (next_out + fifo_out) started to become messy, so I implemented option 2 for now, but left a note in the code so we come back to this in the future if need be.

### DIFF
--- a/lib/nx_deflate.c
+++ b/lib/nx_deflate.c
@@ -838,6 +838,17 @@ static uint32_t nx_compress_nxstrm_to_ddl_out(nx_streamp s)
 	return total;
 }
 
+/* spbc is found in two possible locations Table 6-5 */
+static int get_spbc(nx_streamp s, int fc) {
+	if (fc == GZIP_FC_COMPRESS_RESUME_DHT_COUNT ||
+	    fc == GZIP_FC_COMPRESS_RESUME_FHT_COUNT ||
+	    fc == GZIP_FC_COMPRESS_DHT_COUNT        ||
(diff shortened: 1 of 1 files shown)


lib/nx_deflate.c
```diff
@@ -1738,14 +1741,14 @@ int nx_deflate(z_streamp strm, int flush)
...
-			s->strategy = Z_FIXED;
+			/* compressed data has expanded; so emit a literal block instead */
+			s->need_stored_block = s->spbc; /* amount to repeat */
 			s->tebc = 0; /* not valid since we're
 				      * repeating; last block would
 				      * have sync flushed */
-			prt_info("%s:%d Expanded, trying fixed Huffman, spbc %d, tebc %d\n", __FUNCTION__, __LINE__, s->spbc, s->tebc);
+			prt_info("%s:%d need_stored_block, spbc %d\n", __FUNCTION__, __LINE__, s->spbc);
+			/* TODO Could we use memcpy to copy input to output in place without
+			   the need to call the engine again with WRAP function? */
```
tuliom, line 1751: Yes. Depending on the length to copy, memcpy might be faster.

tuliom (approved):

rzinsly (approved):

Issue #33, opened 2020-06-25
Type mismatches compared to zlib declarations


pedromfcarvalho:
There are type mismatches relative to the zlib definitions.

I'll list the ones I saw below.

One possible issue with this is what happens when a binary compiled using <zlib.h> runs with libnxz with the zlib interfaces.

- lib/nx_adler32.c

Functions adler32 and adler32_z use const char * for the "buf" argument, but the zlib declaration uses Bytef, which I believe resolves to unsigned char *.

Our implementation nx_adler32_z also takes a const char *.  I'm also worried that this could cause some error in the code for nx_adler32_z, since it does arithmetic manipulation of the values in the buffer, and it is derived from the zlib adler32 function, which expects an unsigned char array.

- Function crc32 (lib/nx_crc.c)

This one uses uint64_t instead of uInt for the len parameter.  The function that it calls is crc32_ppc, which takes an int.

What happens here?  Does the unsinged int from the caller that used <zlib.h> get zero-extended to a uint64_t?

Also, our nx_crc32 takes a uin64_t, and calls nx_crc32_z, which iself takes a z_size_t.  Note that our crc32 doesn't call nx_crc32, it calls crc32_ppc.

Function crc32_combine and crc32_combine64 both use uint64_t instead of z_off_t and z_off64_t, respectively, in the zlib definition.  The type that z_off64_t represents varies according to some preprocessor definitions in zlib.h.  I'm not sure what it resolves to in a regular zlib build for ppc64.

- Declaration qualifiers

Our declarations don't have the same qualifiers as the ones in zlib.h, e.g. ZEXTERN, ZEXPORT, OF, etc.  I'm not sure if these actually resolve to something when you build zlib for ppc64.

mscastanho:
Fixed by #144

PR #144, opened 2022-02-17
Fix type mismatches in several functions

mscastanho:
This addresses #33. Since the ABI was changed, it required regenerating `libnxz.abi`. I tried to make the diff as clean as possible (see first commit), but in the end abidw changed the order in which some tags appear in the final XML, causing a much larger diff than needed. I couldn't find a way to avoid that for now.

See the commit messages for more details on the changes.


lib/nx_crc.c
```diff
@@ -261,7 +261,7 @@ local unsigned long nx_crc32_z(crc, buf, len)
 unsigned long ZEXPORT nx_crc32(crc, buf, len)
     unsigned long crc;
     const unsigned char FAR *buf;
-    uint64_t len;
+    unsigned int len;
```
tuliom, line 264: This function is part of the API, we have to be very careful before changing it.
Why are we using `unsigned int` if other functions use `unsigned long` for the same parameter?

lib/Versions
```diff
@@ -28,6 +28,7 @@ LIBNXZ_0.61.0 {
     nx_adler32_combine;
     nx_adler32_z;
     nx_crc32;
+    nx_crc32_z;
```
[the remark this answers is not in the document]
mscastanho: Because I made a mistake xD. I'll fix that on the next push.

lib/nx_adler32.c
```diff
@@ -80,8 +80,8 @@ typedef size_t  z_size_t;
 /* ========================================================================= */
 unsigned long nx_adler32_z(adler, buf, len)
     unsigned long adler;
-    const char *buf;
-    z_size_t len;
+    const unsigned char *buf;
+    size_t len;
```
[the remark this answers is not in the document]
mscastanho: This file has a lot of code copied directly from zlib, where K&R style is used for the whole lib. I agree we should change the style in a separate PR/commit.

lib/crc32_ppc.c
```diff
@@ -42,11 +42,11 @@ static unsigned int crc32_align(unsigned int crc, unsigned char *p,
 }
 #endif
 
-unsigned int __crc32_vpmsum(unsigned int crc, unsigned char *p, 
+unsigned int __crc32_vpmsum(unsigned int crc, const unsigned char *p,
                             unsigned long len);
 
-static uint32_t crc32_vpmsum(uint32_t crc, unsigned char *p,
-                             unsigned len)
+static unsigned long crc32_vpmsum(unsigned int crc, const unsigned char *p,
```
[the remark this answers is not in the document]
mscastanho: Right, I had "make everything match crc32's types" in mind, but I agree it's not necessary for the internal functions. I'll remove these changes on the next push.

lib/crc32_ppc.c
```diff
@@ -80,19 +80,19 @@ static uint32_t crc32_vpmsum(uint32_t crc, unsigned char *p,
...
-uint32_t crc32_ppc(uint32_t crc, unsigned char *data, unsigned len) {
-  if (!data) {
-    return 0;
-  } else {
-    crc = crc32_vpmsum(crc, data, (unsigned long)len);
-  }
-  return crc;
-}
+unsigned long crc32_ppc(unsigned long crc, const unsigned char *data,
+                        unsigned long len) {
```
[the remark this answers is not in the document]
mscastanho: Answered in the comment above.

lib/crc32_ppc.c
```diff
@@ -80,19 +80,19 @@ static uint32_t crc32_vpmsum(uint32_t crc, unsigned char *p,
...
+  if (!data)
+      return 0;
 
+  /* zlib uses unsigned long for crc32, even though it's just 32 bits long, so
+     the truncation should not cause any data loss. */
+  return crc32_vpmsum((unsigned int) (crc & 0xffffffffUL), data, len);
```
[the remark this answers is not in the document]
mscastanho: Agreed. This will be removed on next push.

lib/nx_crc.c
tuliom: @mscastanho based on your explanation, zlib made a mistake and corrected it later with the `*_z` functions.
IMHO, we don't need to make the same mistake for the `nx_*` functions.
However, it's very important that the functions exported by zlib do have the same declaration and definition.

Besides that, I think we can fix the length for `nx_adler32`. Luckily, a 64-bit unsigned integer is binary compatible with a 32-bit unsigned integer.

tuliom: Furthermore: the `nx_*` API is ours to choose.  We can create more functions as we wish.

mscastanho: @tuliom I changed both `nx_adler32` and `nx_crc32` to use `size_t` for the len parameter. So now they are basically equivalent to their `_z` counterparts`.

lib/Versions
```diff
@@ -57,6 +57,7 @@ LIBNXZ_0.61.0 {
 };
 
 LIBNXZ_0.63.0 {
+    nx_crc32_z;
```
tuliom, line 60: what is the goal of exporting `nx_crc32_z` if it's identical to `nx_crc32`?

libnxz.h
```diff
@@ -50,12 +50,13 @@
...
-extern ulong nx_adler32(ulong adler, const char *buf, uint len);
-extern ulong nx_adler32_z(ulong adler, const char *buf, size_t len);
+extern ulong nx_adler32(ulong adler, const unsigned char *buf, size_t len);
+extern ulong nx_adler32_z(ulong adler, const unsigned char *buf, size_t len);
 extern ulong nx_adler32_combine(ulong adler1, ulong adler2, off_t len2);
-extern ulong nx_crc32(ulong crc, const unsigned char *buf, uint64_t len);
-extern ulong nx_crc32_combine(ulong crc1, ulong crc2, uint64_t len2);
-extern ulong nx_crc32_combine64(ulong crc1, ulong crc2, uint64_t len2);
+extern ulong nx_crc32(ulong crc, const unsigned char *buf, size_t len);
+extern ulong nx_crc32_z(ulong crc, const unsigned char *buf, size_t len);
```
tuliom, line 57: Likewise here. Same parameters in both `nx_crc32` functions.

libnxz.h
```diff
@@ -50,12 +50,13 @@
 typedef unsigned int uInt;
 
 /* zlib crc32.c and adler32.c */
-extern ulong nx_adler32(ulong adler, const char *buf, uint len);
-extern ulong nx_adler32_z(ulong adler, const char *buf, size_t len);
+extern ulong nx_adler32(ulong adler, const unsigned char *buf, size_t len);
+extern ulong nx_adler32_z(ulong adler, const unsigned char *buf, size_t len);
```
tuliom, line 54: Both `nx_adler32` functions have the same parameters.
Unfortunately, `nx_adler32_z` has been around since 0.61 and we can't remove it from the ABI.
However, we should be able to remove it from the API here.

tuliom (changes requested):

lib/Versions
mscastanho: Now that `nx_crc32` and `nx_crc32_z` are the same, there's really no point. I'll remove it.

libnxz.h
mscastanho: Done. I removed it from the API but left the function definition in the `nx_adler32.c`. I also made `nx_adler32` contain the actual function implementation and `nx_adler32_z` is just an alias to it now.

tuliom (approved):

PR #145, opened 2022-02-18
Split the selector modes for delfate and inflate

rzinsly:


test/nx-zlib.conf
```diff
@@ -43,6 +43,14 @@ logfile = ./nx.log
 # default: 0
 #nx_selector = 0
 
+# same as nx_selector (0,1,2) for compression, is superseeded by nx_selector.
+# default: 0
+#comp_strategy = 0
+
+# same as nx_selector (0,1,2) for decompression, is superseeded by nx_selector.
```
mscastanho, line 50: Not sure if `superseeded` is the best word here. What about this instead?
```
# same as nx_selector (0,1,2) for decompression. Is ignored if nx_selector is also set.
```

mscastanho (changes requested):
Code looks good, but I have some suggestions on the wording.

`deflate_strategy` is not the best nomenclature IMHO, because it's already used by zlib for something else (used to tune the compression algorithm). What about `deflate_mode` and `inflate_mode` instead?

lib/nx_zlib.h
```diff
@@ -151,7 +156,8 @@ struct nx_config_t {
 	int      strategy_override; /** Force use of an specific deflate
 				     * strategy.  0 is fixed huffman, 1 is
 				     * dynamic huffman */
-	uint8_t  gzip_selector;
+	struct selector st_sel; /** strategy selector: selects between
```
mscastanho, line 159: Now that we're using `{deflate,inflate}_mode`, this should be changed as well. To `mode_selector` or something similar. Same for the comment.

mscastanho (changes requested):

lib/nx_zlib.h
rzinsly: @mscastanho How about calling it `mode` and removing the `_mode` from  `{deflate,inflate}_mode`?

mscastanho: @rzinsly Sounds good.

Issue #29, opened 2020-06-19
Review the usage of crc32_ppc and nx_crc32


tuliom:
`crc32_ppc` provides a vectorized implementation, while `nx_crc32` is more generic.
libnxz needs a CRC32 implementation internally to create header checksum, which is just "a few bytes".
Vectorized code are rarely the fastest implementation for small length.

Can we use the vectorized code in all scenarios even if it isn't the fastest for small length? i.e. the impact would be small.
If not, we may need to implement a smart switch depending on the length.

tuliom:
@abalib just sent me the following regarding the usage of vectorized code for all scenarios:
> I think the answer is yes.   We should test it though.  I can't remember now if there are any gotchas

abalib:
Your idea of using only the vectorized code makes sense to me.  In my 
humble opinion, no need for a smart switch.  This library only runs on 
POWER anyway.

Although, I can't remember now if there are any gotchas with small data. 
Testing should reveal if any.

PR #146, opened 2022-02-22
Replace assembly-based crc32 with builtin-based implementation

mscastanho:
This is based on top of #144, so the first two commits are from there. I'll rebase once it gets merged.

I ran some performance tests with random data from `/dev/random` to help decide if we should use the vector implementation for all lenghts (#29). Turns out the generic implementation is indeed faster up to length 31 (first plot), because for those lenghts the vector implementation relies on successive table lookups to calculate the checksum, no vectors involved. 

image

So I decided to get the best of both worlds and only apply the vector version for length larger than 32:

image


rzinsly (approved):
LGTM

lib/nx_crc.c
```diff
@@ -36,26 +36,17 @@
...
 /* make crc32.c work without external zlib headers */
 #undef MAKECRCH
 #undef DYNAMIC_CRC_TABLE
-#define FAR
 #define local static
-#define OF(args)  ()
-typedef unsigned long uLong;
-typedef unsigned int uInt;
-typedef uint32_t z_crc_t;
 typedef size_t   z_size_t;
```
mscastanho, line 45: I rebased on top of current develop and also re-added this line. Without the project doesn't compile on RHEL7 for the same reason as in 74b61ae

tuliom (approved):

PR #148, opened 2022-03-03
Use history during compression

mscastanho:
This PR adds functionality to maintain and pass history data to NX, which helps achieving better compression ratios. Right now this will be used by default, but we can limit history usage to higher compression levels in the future.

To get it right I had to touch many places in `nx_deflate.c` at once, so the diff is rather large. Unfortunately I wasn't able to split it into separate commits while guaranteeing everything was functional for each commit.

Below is a comparison of compression ratio for all files used by our  `oct` tests. `libnxz` is current `develop` branch, `libnxz w/ history` is this PR and zlib is whatever zlib version available on Debian 11.

I still haven't investigated the impact on overall throughput compression, that's why this is marked as a Draft.

| filename                     | original size |             libnxz |  libnxz w/ history |               zlib |
|--|--:|--:|--:|--:|
   | at.uncompressed              |    2496942080 | 670490807 (26.85%) | 599158145 (24.00%) | 455172323 (18.23%) |
   | calgary.uncompressed         |       3265536 |   1515417 (46.41%) |   1375807 (42.13%) |   1068355 (32.72%) |
   | initrd.uncompressed          |      58841600 |  24007696 (40.80%) |  22678637 (38.54%) |  18398071 (31.27%) |
   | linux.uncompressed           |    1019310080 | 277016818 (27.18%) | 243936993 (23.93%) | 185251455 (18.17%) |
   | Packages.uncompressed        |        216674 |    152991 (70.61%) |    148005 (68.31%) |     94062 (43.41%) |
   | patch.uncompressed           |        949514 |    356942 (37.59%) |    329506 (34.70%) |    269018 (28.33%) |
   | random12k.uncompressed       |         12288 |      4465 (36.34%) |      4465 (36.34%) |      4232 (34.44%) |
   | random13M.uncompressed       |      13631488 | 13637044 (100.04%) | 13794165 (101.19%) | 13635654 (100.03%) |
   | random16k.uncompressed       |         16384 |      8564 (52.27%) |      4556 (27.81%) |      4267 (26.04%) |
   | random4k.uncompressed        |          4096 |     4107 (100.27%) |     4107 (100.27%) |     4107 (100.27%) |
   | silesia-dickens.uncompressed |      10192446 |   4920121 (48.27%) |   4565008 (44.79%) |   3871634 (37.99%) |
   | silesia-mozilla.uncompressed |      51220480 |  24919913 (48.65%) |  22681382 (44.28%) |  19089124 (37.27%) |
   | silesia-mr.uncompressed      |       9970564 |   5080216 (50.95%) |   4925439 (49.40%) |   3675941 (36.87%) |
   | silesia-nci.uncompressed     |      33553445 |   5400767 (16.10%) |   5099454 (15.20%) |    3200188 (9.54%) |
   | silesia-ooffice.uncompressed |       6152192 |   3586821 (58.30%) |   3455765 (56.17%) |   3097294 (50.34%) |
   | silesia-osdb.uncompressed    |      10085684 |   6271468 (62.18%) |   4542795 (45.04%) |   3695170 (36.64%) |
   | silesia-reymont.uncompressed |       6627202 |   2521550 (38.05%) |   2333028 (35.20%) |   1860871 (28.08%) |
   | silesia-samba.uncompressed   |      21606400 |   7554879 (34.97%) |   6871239 (31.80%) |   5451405 (25.23%) |
   | silesia-sao.uncompressed     |       7251944 |   6012875 (82.91%) |   5849831 (80.67%) |   5331799 (73.52%) |
   | silesia-webster.uncompressed |      41458703 |  15896248 (38.34%) |  14678002 (35.40%) |  12213947 (29.46%) |
   | silesia-xml.uncompressed     |       5345280 |   1302018 (24.36%) |   1143531 (21.39%) |    687997 (12.87%) |
   | silesia-x-ray.uncompressed   |       8474240 |   6489959 (76.58%) |   6483337 (76.51%) |   6045117 (71.34%) |
   | sparse1000M.uncompressed     |    1048576000 |   10838372 (1.03%) |   12714697 (1.21%) |    1019185 (0.10%) |
   | sparse10M.uncompressed       |      10485760 |     109412 (1.04%) |     127177 (1.21%) |      10209 (0.10%) |
   | vmlinux.uncompressed         |      21982000 |   9087429 (41.34%) |   8589303 (39.07%) |   7407614 (33.70%) |
   | zero13M.uncompressed         |      13631488 |     141924 (1.04%) |     165321 (1.21%) |      13269 (0.10%) |
   | zero4k.uncompressed          |          4096 |         80 (1.95%) |         80 (1.95%) |         26 (0.63%) |

### DIFF
--- a/lib/nx_deflate.c
+++ b/lib/nx_deflate.c
@@ -630,9 +630,6 @@ int nx_deflateInit2_(z_streamp strm, int level, int method, int windowBits,
 		return Z_STREAM_ERROR;
 	}
 
-	/* only support level 6 here */
-	level = 6;
-
 	s = nx_alloc_buffer(sizeof(*s), nx_config.page_sz, nx_config.mlock_nx_crb_csb);
 	if (s == NULL) return Z_MEM_ERROR;
 	memset(s, 0, sizeof(*s));
@@ -649,6 +646,34 @@ int nx_deflateInit2_(z_streamp strm, int level, int method, int windowBits,
 		s->strategy = Z_FIXED;
 	else
 		s->strategy = Z_DEFAULT_STRATEGY;
+
+	switch (level) {
+		case 0:
+			/* TODO: Level 0 is not working properly */
+			s->level = 6;
+			break;
+		case Z_DEFAULT_COMPRESSION:
+		case 1:
+		case 2:
+		case 3:
+		case 4:
+			s->max_history_len = 0;
+			break;
+		case 5:
+		case 6:
+		case 7:
+			/* From level 5 upwards we'll use different history lengths, starting
+			   from 4 kB. */
+			s->max_history_len = 1<<(level+7);
+			break;
+		case 8:
+		case 9:
+			s->max_history_len = DEF_HIST_LEN;
+			break;
+		default:
+			/* Invalid level */
+			return Z_STREAM_ERROR;
+	}
 
 	s->zstrm      = strm; /* pointer to parent */
 	s->page_sz    = nx_config.page_sz;
@@ -755,11 +780,11 @@ static inline void cache_input(nx_streamp s)
 		s->len_in/2 - s->cur_in - s->used_in : 0;
 
 	/* Not enough space to cache input. Time to throw away old data, keeping at
-	 * most DEF_HIST_LEN bytes of history. */
+	 * most s->max_history_len bytes of history. */
 	if(free_bytes < s->avail_in) {
-		memmove(s->fifo_in, s->fifo_in + s->cur_in - DEF_HIST_LEN,
-				DEF_HIST_LEN + s->used_in);
-		s->cur_in = DEF_HIST_LEN;
+		memmove(s->fifo_in, s->fifo_in + s->cur_in - s->max_history_len,
+				s->max_history_len + s->used_in);
+		s->cur_in = s->max_history_len;
 		free_bytes = s->len_in/2 - s->cur_in - s->used_in;
 	}
 	copy_bytes = NX_MIN(free_bytes, s->avail_in);
@@ -985,10 +1010,10 @@ static int  nx_compress_block_update_offsets(nx_streamp s, int fc)
 	   need to throw away old data if we reach that limit. */
 	if (s->cur_in + from_next_in > s->len_in/2) {
 
-		/* Throwing away old data. We just want to keep DEF_HIST_LEN bytes of
+		/* Throwing away old data. We just want to keep s->max_history_len bytes of
 		   history at the beginning of fifo_in. Data from next_in has priority
 		   because it comes later in the data stream. */
-		copy_bytes = from_next_in < DEF_HIST_LEN ? DEF_HIST_LEN-from_next_in : 0;
+		copy_bytes = from_next_in < s->max_history_len ? s->max_history_len-from_next_in : 0;
 		if (copy_bytes > 0)
 			memmove(s->fifo_in, s->fifo_in + s->cur_in - copy_bytes, copy_bytes);
 		s->cur_in = copy_bytes;
@@ -996,13 +1021,13 @@ static int  nx_compress_block_update_offsets(nx_streamp s, int fc)
 
 	/* Copy enough bytes from the end of the input data to fill the history */
 	if (from_next_in > 0) {
-		copy_bytes = NX_MIN(from_next_in, DEF_HIST_LEN);
+		copy_bytes = NX_MIN(from_next_in, s->max_history_len);
 		memcpy(s->fifo_in + s->cur_in,
 			   s->next_in + from_next_in - copy_bytes, copy_bytes);
 		s->cur_in += copy_bytes;
 	}
 
-	s->history_len = NX_MIN(s->cur_in, DEF_HIST_LEN);
+	s->history_len = NX_MIN(s->cur_in, s->max_history_len);
 
 	update_stream_in(s, from_next_in);
 	update_stream_in(s->zstrm, from_next_in);
@@ -2036,7 +2061,7 @@ int nx_deflateSetDictionary(z_streamp strm, const unsigned char *dictionary, uns
 	s->dict_len = dictLength;
 	s->dict_id = adler;
 	s->cur_in += left;
-	s->history_len = NX_MIN(s->history_len + dictLength, DEF_HIST_LEN);
+	s->history_len = NX_MIN(s->history_len + dictLength, s->max_history_len);
 
 	/* Mimic zlib behavior */
 	s->zstrm->total_in += NX_MIN(dictLength, DEF_MAX_DICT_LEN);
--- a/lib/nx_deflate.c
+++ b/lib/nx_deflate.c
@@ -569,7 +569,6 @@ int nx_deflateEnd(z_streamp strm)
 
 	nx_free_buffer(s->fifo_in, s->len_in, 0);
 	nx_free_buffer(s->fifo_out, s->len_out, 0);
-	nx_free_buffer(s->dict, s->dict_alloc_len, 0);
 
 	nx_close(s->nxdevp);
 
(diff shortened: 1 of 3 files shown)


abalib:
Looks pretty good.  Picked up about 3% for many.  Osdb shows the best improvement.   Compression is done in 1MB chunks I think.  Is your patch populating the resume buffer with the last 32KB of the previous 1MB?

mscastanho:
@abalib the patch keeps the last 32 kB of input in `fifo_in` at all times, and that is passed to NX as history data. So if the user gives us 1MB at a time, the 32 kB of history will be the last bytes of the previous chunk. But if data is given is smaller chunks (say 4096), it will correspond to data from several previous jobs.

lib/nx_deflate.c
```diff
@@ -745,16 +743,25 @@ static int nx_copy_fifo_out_to_nxstrm_out(nx_streamp s)
...
+	/* Avoid underflow if more than half of fifo_in has been used already. */
+	free_bytes = s->cur_in + s->used_in < s->len_in/2 ?
+		s->len_in/2 - s->cur_in - s->used_in : 0;
 
-	free_bytes = s->len_in/2 - s->cur_in - s->used_in;
+	/* Not enough space to cache input. Time to throw away old data, keeping at
+	 * most DEF_HIST_LEN bytes of history. */
+	if(free_bytes < s->avail_in) {
+		memmove(s->fifo_in, s->fifo_in + s->cur_in - DEF_HIST_LEN,
+				DEF_HIST_LEN + s->used_in);
```
tuliom, line 761: If we're reaching this point, that means the request had a small number of bytes.
Is it really worth doing this here?
See my next comment.

lib/nx_deflate.c
```diff
@@ -933,10 +971,38 @@ static int  nx_compress_block_update_offsets(nx_streamp s, int fc)
...
+	   also need to limit the amount of data in fifo_in to len_in/2, so we may
+	   need to throw away old data if we reach that limit. */
+	if (s->cur_in + from_next_in > s->len_in/2) {
+
+		/* Throwing away old data. We just want to keep DEF_HIST_LEN bytes of
+		   history at the beginning of fifo_in. Data from next_in has priority
+		   because it comes later in the data stream. */
+		copy_bytes = from_next_in < DEF_HIST_LEN ? DEF_HIST_LEN-from_next_in : 0;
+		if (copy_bytes > 0)
+			memmove(s->fifo_in, s->fifo_in + s->cur_in - copy_bytes, copy_bytes);
```
tuliom, line 993: At the risk of suggesting a premature optimization...
I believe this copy could also make room for `nx_config.cache_threshold` bytes.
This would guarantee that `cache_input()` would always have enough room for itself.

tuliom (commented):
LGTM.
I think it's just missing some performance data just to guarantee this can indeed be used for level = 6 or if it should require a higher level.

mscastanho:
Here goes some performance data using the files from `oct`. I tested a few history lengths and compared the time to compress each file and the final compressed size. Note all values are relative to what we get when using `develop`.

all

Looks like using the full 32 kB of history can be quite expensive. The penalty was >20% for most files, while the extra compression did not follow the same progression. So I think using 32 kB of history may be too expensive for the default compression level. It's probably time to start tuning the compression levels.

Here's my first suggestion for compression level ->  history length mapping:
```
level 0 - 4 : no history
level 5 : 4 kB of history
level 6 : 8 kB of history
level 7 : 16 kB of history
level 8 : 32 kB of history
level 9 : 32 kB of history
```
My reasoning is as follows: 
1. we should gradually increase the history length for higher levels to provide some balance between better compression and reasonable compression time.
2. level 9 (i.e. the highest) should not be the first level to use the entire history. There are other techniques we can still use to get better compression (like doing a second pass with LZ77 symbols from first pass) that may also increase compression time significantly, so it would be nice if we could give the user the choice to use as much history as possible (level 8), but without using other techniques (which would be left only to level 9).

**note**: I'm still unsure about what's happening with `silesia-mozilla` and `silesia-samba`. I ran the benchmark several times and using some history (4k, 8k) is always faster than using no history at all (`develop`), which doesn't make much sense to me.

lib/nx_deflate.c
mscastanho: I experimented with this, but couldn't get it to pass all tests. I'll give a second shot soon.

abalib:
>Looks like using the full 32 kB of history can be quite expensive. The penalty was >20% for most files, 
I intuit that 20% penalty cannot be right.  Because 32KB/1MB is only 3% extra data going in to the hardware. This is relative to the no-history case.   Although, if you are copying 32KB from input buffer to the history buffer (e.g., with memcpy), then I believe 20% might be possible.  In any case, it is what it is.  Great investigation and work...

mscastanho:
> if you are copying 32KB from input buffer to the history buffer (e.g., with memcpy)
This is indeed happening. We need these copies to keep the history buffer consistent. But it's not copying 32 kB for every NX job.

Also, I'm afraid we're not sending 1MB of data to the hardware on each job. I used `minideflate` as the application to run these tests. By default it uses a 4 kB buffer to pass data to zlib/libnxz. Our default `cache_threshold` is 8 kB. So combining these two values, an NX job will be issued when the amount of input (cached + `next_in`) hits 12 kB (first time it's over threshold), which ends up being the amount of data given to the hardware on each job.

In this 12 kB job, 8 kB of data comes from `fifo_in` (cached, already memcpy'd, same as on `develop`), and 4 kB from the user (not cached). When using the history these extra 4 kB in `next_in` will be copied to `fifo_in` as well. Also, once `fifo_in` gets half full (every 64 kB), we'll do a 32 kB copy to reset the history buffer.

So if we were to compare only number of bytes copied with `memcpy` before and after this patch, we're likely increasing it 2x or more. But when we factor in the other costs (sanity checks, lots of branches, job submission, result processing, etc) it boils down to ~20-40% overall penalty (based on the results above).

We may actually be able to alleviate that penalty by increasing `cache_threshold` (to reduce copies from `next_in` to `fifo_in` to compose history) and `s->len_in` (to reduce the number of `fifo_in` resets we need to make, and so the number of copies).

lib/nx_deflate.c
mscastanho: @tuliom I'm working on an optimization that completely removes these `memmove` calls to reset `fifo_in`. But it still has some bugs, so I'll probably include it in a separate PR.

> I believe this copy could also make room for nx_config.cache_threshold bytes.
Note this copy already makes as much room as possible on `fifo_in`. We only copy the amount of data needed to maintain the history, so at most 32 kB.

lib/nx_deflate.c
```diff
@@ -630,9 +630,6 @@ int nx_deflateInit2_(z_streamp strm, int level, int method, int windowBits,
 		return Z_STREAM_ERROR;
 	}
 
-	/* only support level 6 here */
-	level = 6;
```
rzinsly, line 628: You are enabling all levels here but level 0 is not working well at the moment and needs to be fixed.
I haven't tested the others extensively but they seem to work.

lib/nx_deflate.c
```diff
@@ -650,6 +647,15 @@ int nx_deflateInit2_(z_streamp strm, int level, int method, int windowBits,
 	else
 		s->strategy = Z_DEFAULT_STRATEGY;
 
+	if (level < 5 || level > 9)
```
rzinsly, line 650: I think we shouldn't accept `(level < 0 || level > 9)`.

rzinsly (changes requested):

tuliom (approved):
Considering the following will be left for a future work:
1. Memory copy optimization.
2. Increase of compression tests in oct for all levels.

This LGTM.
Notice there are some important issues pointed out by Raphael.

lib/nx_deflate.c
mscastanho: Agreed. I replaced the ifs with a switch-case. Now all these values whould return `Z_STREAM_ERROR` like zlib does.

mscastanho:
Rebased on top of the current develop branch.

I added a switch-case to handle the levels now, making sure it fails for <0 and >9 and I also forced `Z_DEFAULT_COMPRESSION` to not using any history (as before), so current users should not see any undesired performance regressions for the default case.

rzinsly (approved):

mscastanho:
@tuliom I created #156 and #157 to keep track of the two items you mentioned

PR #149, opened 2022-03-08
Remove `ZLIB_API`

tuliom:
When the zlib API is removed from the library, the generated shared
object becomes binary incompatible with zlib and with a libnxz that
supports the zlib API.  This scenario should be avoided.

The support for dynamic switching between hardware and software
implementations require the zlib API to always be present.

### DIFF
--- a/configure
+++ b/configure
@@ -815,7 +815,6 @@ with_gzip
 with_sha256sum
 with_valgrind
 with_wget
-enable_zlib_api
 enable_doxygen_doc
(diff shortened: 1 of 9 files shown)


rzinsly (approved):

tuliom:
I rebased the patch on top of the develop branch.

mscastanho (approved):

tuliom:
Fixed another minor conflict.

PR #150, opened 2022-03-21
inflate: Limit the amount of pages touched

tuliom:
Before inflating a file, all the input and output pages are touched.
However, when this file is too large, there is no need to touch all its
pages, just as many as we can fit in a single NX job.

By not doing this, we end up dealing with too many page faults than
necessary for each NX request.


tuliom:
Let me clarify: I'm rewriting this PR and I'm changing exactly how `target_sz` and `target_sz_expected` are affecting the algorithm.

tuliom:
I have incorporated @abalib 's comments. However, while looking at the code I found an issue (commit `Fix the total amount of pages being touched `) and I realized it was necessary to change the code that manipulate the DDE by limiting the amount of input/output sent to the NX instead of just limiting the amount of pages touched.

In other words: if we send too much input/output, we have to touch all of their pages, otherwise we'll have to pay the price for treating `ERR_NX_AT_FAULT`.
If we send too little, we decrease performance.
So, I decided to use the values in `source_sz_expected` and `target_sz_expected` plus an increase in the amount of output whenever that's available in order to avoid excessive `ERR_NX_DATA_LENGTH`.

Unfortunately this PR has now 6 commits.
Commits 2, 4 and 5 are just reorganizing/improving code.
I suggest reviewers to take a look one commit at a time.

I also think this is ready for review now.

abalib (commented):
I trust that buf_len cannot go negative here https://github.com/libnxz/power-gzip/pull/150/commits/601f718fdc0b64e58669b114c9e0d33c218f87f9

abalib:
> So, I decided to use the values in `source_sz_expected` and `target_sz_expected` plus an increase in the amount of output whenever that's available in order to avoid excessive `ERR_NX_DATA_LENGTH`.

Looks right conceptually

lib/nx_inflate.c
```diff
@@ -1261,30 +1382,34 @@ static int nx_inflate_(nx_streamp s, int flush)
...
-			source_sz = source_sz + nx_history_len;
-
-			if (target_sz > (2 * INF_MAX_EXPANSION_BYTES))
-				target_sz = (target_sz + 1) / 2;
-			else if (target_sz > INF_MAX_EXPANSION_BYTES)
-				target_sz = INF_MAX_EXPANSION_BYTES;
+			source_sz_expected = source_sz - nx_history_len;
+			if (source_sz_expected > (2 * INF_MIN_INPUT_LEN))
+				source_sz_expected = (source_sz_expected + 1)
+						     / 2;
```
mscastanho, line 1388: Please put this in the previous line, it's easier to read.

lib/nx_inflate.c
```diff
@@ -1261,30 +1382,34 @@ static int nx_inflate_(nx_streamp s, int flush)
...
+			else if (source_sz_expected > INF_MIN_INPUT_LEN)
+				source_sz_expected = INF_MIN_INPUT_LEN;
+
+			/* else if caller gave fewer source bytes, keep it as
+			   is.  */
+			source_sz = source_sz_expected + nx_history_len;
+
+			if (target_sz_expected > (2 * INF_MAX_EXPANSION_BYTES))
+				target_sz_expected = (target_sz_expected + 1)
+						     / 2;
```
mscastanho, line 1398: Please put this in the previous line, it's easier to read.

mscastanho (changes requested):
Code looks good and I can verify the huge performance improvement on PowerVM using `compdecomp_th`. Nice work!

I have just some minor suggestions. With those changes LGTM.

lib/nx_inflate.c
tuliom: @mscastanho I had to break the line because it was too long. Any other suggestions?
What do you think about this?
```
source_sz_expected
        = (source_sz_expected + 1)  / 2;
```

lib/nx_inflate.c
tuliom: Likewise here.

lib/nx_inflate.c
mscastanho: @tuliom that's better. 

Are you using 8 spaces to visualize tabs? I think my editor is using 4, that's why the line didn't look too long to me.

tuliom: Yes

tuliom:
@mscastanho I updated the patches with the fixes that you requested.

mscastanho (approved):

lib/nx_inflate.c
```diff
@@ -1280,8 +1301,11 @@ static int nx_inflate_(nx_streamp s, int flush)
 
 	/* fault in pages */
 	nx_touch_pages_dde(ddl_in, source_sz, nx_config.page_sz, 0);
-	nx_touch_pages_dde(ddl_out, target_sz, nx_config.page_sz, 1);
```
rzinsly, line 1223: This line doesn't need changes, but I'm ok with it.

rzinsly (approved):

lib/nx_inflate.c
tuliom: Good point. A previous change didn't work well and I forgot to revert the indentation changes.

PR #158, opened 2022-04-12
Convert .abi files to libabigail 2.0 format

mscastanho:
libabigail 2.0.0 introduced a new incompatible XML format for the ABI
description, so we can't use files generated with prior versions to
compare with a newer one. To avoid issues, convert both libnxz.abi and
libz.abi to vesion 2.0 and skip test_abi if libabigail is older than
2.0.0.


tuliom:
This PR conflicts with PR #138 . It's important to synchronize both before merging them.

mscastanho:
Rebased on top of `develop` to fix conflicts with #138

rzinsly (approved):

Issue #161, opened 2022-04-20
oct: Regression on gzip tests


tuliom:
I started seeing some gzip-related failures when running oct tests recently. I can reproduce these failures without testing with large files.

After bisecting, commit 764fe41e9a232d9b724bd1760b3b941c9cdc2387 seems to be the culprit.

rzinsly:
Can you reproduce it with 1e335c4? Seems the lack of `gzread` on 764fe41 cause other failures.

PR #167, opened 2022-05-04
Fix issue 161

tuliom:
Fix issue #161 by enforcing `ERR_NX_TPBC_GT_SPBC` on all expected cases.

Also improve the `gzwrite()`'s execution by processing more input in case libnxz output cache has too much data.

### DIFF
--- a/lib/nx_gzlib.c
+++ b/lib/nx_gzlib.c
@@ -223,28 +223,41 @@ int nx_gzwrite (gzFile file, const void *buf, unsigned len)
 	z_streamp stream = &(state->strm);
 	int rc;
 	uLong last_total_in = stream->total_in;
-	unsigned char* next_out = malloc(sizeof(char)*len);
-
+	unsigned char* next_out = malloc(len);
 	if (next_out == NULL)
 		return Z_MEM_ERROR;
 
+	/* Under some scenarios (e.g. when compressing random data), the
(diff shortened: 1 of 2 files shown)


lib/nx_gzlib.c
```diff
@@ -223,28 +223,41 @@ int nx_gzwrite (gzFile file, const void *buf, unsigned len)
...
+		rc = nx_deflate(stream, Z_NO_FLUSH);
+		ssize_t w_ret = write(state->fd, next_out,
+				      len - stream->avail_out);
+
+		if (w_ret < 0) {
+			free(next_out);
+			state->err = Z_ERRNO;
+			return 0;
+		}
+		if (rc != Z_OK && rc != Z_STREAM_END){
```
mscastanho, line 253: @tuliom what if `nx_deflate` returns `Z_BUF_ERROR` because it generated more data than could fit into `next_out`? In this case `nx_deflate` should be called again with more output space just to flush the cached output data.

rzinsly: @mscastanho doesn't the data go to the `fifo_out` when it cannot fit on `next_out`?
Deflate returns `Z_BUF_ERROR` when you call it without output data, in this case it will always have `len`.

mscastanho: True. Just ignore my previous comment.

mscastanho (approved):

PR #168, opened 2022-05-05
DLPAR support

mscastanho:
The PR is split into many small and self-contained commits to easy the review process. This is marked as draft because there are several pieces still missing, but I'd like to start the discussion soon so we can iron out some details before I delve into the rest of the patch set.

Missing pieces:
- [x] ~`nx_*` functions are not falling back to software compression when window can't be allocated. We discussed this could be desirable if we're in mode `AUTO` or `MIX`, but today these functions are completely agnostic to the modes. This will likely require making most `nx_*` functions into wrappers (like their zlib interface counterparts) and taking into account that they may be using software rather than NX (as most `nx_*` functions assume today). So I'd like to confirm with you that's what we want before I start working on these changes.~
- [x] Currently when a window gets deactivated and we suspect (it's hard to be sure, see comment in `gzip_vas.c`), it just fails and returns to the user. Ideally we should fallback to software, but I'm still unsure this is possible mid-stream, and requires more research and work. Probably as separate PR.
- [x] Improvements to `test_exhaust_credits`: this consumes all credits in the system in order to test libnxz's behavior, so it's not suitable to be run in parallel with other tests or even by default on `make check`. I'm still trying to find a way to get it built but not run by default. Also, it needs to be expanded to test the `nx_*` interface fallback to zlib when that's implemented.
- [x] Policies to help balance window usage: voluntarily closing windows when running out of credits, limiting the amount of credits used by libnxz, etc.

Other points for discussion:
- [ ] When a window can't be allocated due to lack of credits, this first version waits and retries the allocation 10 times (somewhat arbitrary value). The amount of times to retry and ticks to wait between cycles is debatable.


tuliom:
The first 4 commits in this PR LGTM and could be merged before we complete the review of the last 5 commits.

lib/nx_deflate.c
```diff
@@ -2231,27 +2232,39 @@ int deflateInit2_(z_streamp strm, int level, int method, int windowBits,
 			prt_info("this stream has been initialized by sw\n");
 		}
 
-		rc = nx_deflateInit2_(strm, level, method, windowBits, memLevel, strategy, version, stream_size);
-		if(rc != Z_OK){
+		rc = nx_deflate_init(strm, level, method, windowBits, memLevel,
```
tuliom, line 2235: Where is `nx_deflate_init` defined?

tuliom (changes requested):
@mscastanho Regarding `nx_*` functions: I believe it's just a matter of agreement.
They can continue to be used as NX-only functions, returning errors to their callers.
That way, the zlib API would have to get these errors and fallback to zlib, if necessary.

Would this be doable?

lib/nx_inflate.c
```diff
@@ -2021,6 +2021,7 @@ int inflateInit2_(z_streamp strm, int windowBits, const char *version, int strea
 
 		rc = nx_inflateInit2_(strm, windowBits, version, stream_size);
 		if(rc != Z_OK){
+			strm->stage = temp;
```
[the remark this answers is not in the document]
mscastanho: Thanks, will be fixed in the next push.

lib/nx_deflate.c
mscastanho: It was part of another patch that I did not include in this first draft version. I'll fix that on the next push.

mscastanho:
> Regarding nx_* functions: I believe it's just a matter of agreement.
They can continue to be used as NX-only functions, returning errors to their callers.
That way, the zlib API would have to get these errors and fallback to zlib, if necessary.

I agree. I think this behavior makes more sense.

>Would this be doable?

Yes, and it should be mostly implemented in the PR already. Let me double check.

test/test_exhaust_credits.c
```diff
@@ -0,0 +1,187 @@
...
+
+	/* VAS window allocation can be slow on PowerVM, so give it some time */
+	usleep(100);
+
+	if(read_credits(&total_credits, &used_credits))
+		return TEST_ERROR;
+	fprintf(stderr, "Credits after:  total: %d  used: %d\n", total_credits,
+		used_credits);
+
+	generate_random_data(src_len);
```
rzinsly, line 138: Would be good to test if `used_credits >= total_credits` here.

rzinsly:
>nx_* functions are not falling back to software compression when window can't be allocated. We discussed this could be desirable if we're in mode AUTO or MIX, but today these functions are completely agnostic to the modes. This will likely require making most nx_* functions into wrappers (like their zlib interface counterparts) and taking into account that they may be using software rather than NX (as most nx_* functions assume today). So I'd like to confirm with you that's what we want before I start working on these changes.

I don´t know if we really need wrappers, what would they do exactly?

rzinsly:
>Currently when a window gets deactivated and we suspect (it's hard to be sure, see comment in gzip_vas.c), it just fails and returns to the user. Ideally we should fallback to software, but I'm still unsure this is possible mid-stream, and requires more research and work. Probably as separate PR.

I still think it's not feasible to fallback to software mid-stream, IMO unfortunately we don't have much options here.

lib/nx_deflate.c
```diff
@@ -2231,27 +2232,39 @@ int deflateInit2_(z_streamp strm, int level, int method, int windowBits,
...
-
 	}else if(nx_config.mode.deflate == GZIP_NX){
-		rc = nx_deflateInit2_(strm, level, method, windowBits, memLevel, strategy, version, stream_size);
-	}else{
-		rc = sw_deflateInit2_(strm, level, method, windowBits, memLevel, strategy, version, stream_size);
+		rc = nx_deflate_init(strm, level, method, windowBits, memLevel,
+				     strategy, version, stream_size);
+		if (rc == Z_STREAM_ERROR && errno == EAGAIN)
+			/* Failed due to lack of credits on POWERVM  */
+			prt_info("Falling back to software compression\n");
```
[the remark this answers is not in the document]
mscastanho: @rzinsly We can't afford that. This is the zlib interface, the app has no knowledge at all about libnxz, and most apps will most likely have been written with zlib in mind. We can't fail here.

mscastanho:
> I don´t know if we really need wrappers, what would they do exactly?
@rzinsly we would have to add logic to fallback to zlib in case the NX allocation fails. Today most nx_ functions simply assume we just want to use NX and nothing else. So we would have to encapsulate all the existing logic (that uses NX) into a clause of an if statement, that's why I think wrappers would be needed.

But I still think letting the nx_ interface fail if NX is not available makes more sense. If the user wants the fallback behavior they can use the zlib interface.

lib/nx_deflate.c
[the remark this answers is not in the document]
rzinsly: @mscastanho if the mode is set to GZIP_NX the user have knowledge about libnxz, if it's used by another app that fails it's user error, if we use sw when the user explicitly asked for NX seems a library fault to me.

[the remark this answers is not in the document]
tuliom: Good point.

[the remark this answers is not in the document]
mscastanho: @rzinsly You have a point. I changed the code to reflect that.

mscastanho:
@tuliom @rzinsly I made quite a few changes to the PR, including your suggestions. All tests are passing, but I still have to test the behavior when running out of credits, so I'm still leaving this as a draft. Feel free to take a look and suggest changes if you want though.

The number of commits is starting to get big. The first 6 commits are fairly self-contained, so I can put them in a separate PR to facilitate reviewing if you like.

rzinsly:
> The first 6 commits are fairly self-contained, so I can put them in a separate PR to facilitate reviewing if you like.

Personally I do prefer that.

mscastanho:
@tuliom @rzinsly I pushed my latest changes and rebase on top of develop. It may still need some adjustments, but I think it's in a reviewable state and I'd like your feedback on the changes so far.

lib/gzip_vas.c
```diff
@@ -167,12 +167,18 @@ int nx_function_end(nx_devp_t nxhandle)
...
+		if (rc < 0) {
+			fprintf(stderr, "munmap() failed, errno %d\n", errno);
+			return rc;
+		}
+
+		rc = close(nxhandle->fd); /* see issue 164 comment */
+		if (rc < 0) {
+			fprintf(stderr, "close() failed, errno %d\n", errno);
+			return rc;
+		}
```
rzinsly, line 180: Wrong indentation.

lib/gzip_vas.c
```diff
@@ -167,12 +167,18 @@ int nx_function_end(nx_devp_t nxhandle)
...
+		rc = munmap(nxhandle->paste_addr - 0x400, 4096);
+		if (rc < 0) {
+			fprintf(stderr, "munmap() failed, errno %d\n", errno);
+			return rc;
+		}
+
+		rc = close(nxhandle->fd); /* see issue 164 comment */
+		if (rc < 0) {
+			fprintf(stderr, "close() failed, errno %d\n", errno);
+			return rc;
```
rzinsly, line 179: Not an issue but this `return` could be removed.

test/test_exhaust_credits
```diff
@@ -0,0 +1,41 @@
...
+}
+
+echo "=== Testing zlib API ==="
+# Return code 0 means no failures should occur
+test_mode ./exhaust_credits "AUTO" 0 0
+
+# Return code 1 means the number of used should not be N-1
+test_mode ./exhaust_credits "SW" 1 1
+
+# Return code 18 means deflateInit and inflateInit
```
rzinsly, line 29: deflateInit and inflateInit **should fail**

test/test_window_suspend.c
```diff
@@ -0,0 +1,186 @@
+/* Test libnxz behavior when windows are suspended due to a DLPAR operation.
+
+   Test outline:
+   - Consume N-1 credits by calling deflateInit from N-1 child processes
+   - Force a DLPAR core remove operation to remove 3 credits form the system
```
rzinsly, line 5: s/form/from/

test/test_window_suspend.c
```diff
@@ -0,0 +1,186 @@
...
+/* TODO: calling system with super-user privileges is not recommended, implement
+   our own version using fork, exec and wait instead. */
+int run(const char *command)
+{
+	return system(command);
+}
+
+/* For some reason the number of credits only changes when we alter the number
+   of virtual processors, even though on shared LPAR they are tied to the number
+   o processing units (aka Entitled Capactity). Remember credits=proc_units*20,
```
rzinsly, line 55: **of** processing units

rzinsly (commented):
The code is looking good besides some style and comments issues.
I'm not a fan of having a test that needs to be run with root privileges but I can't think in a way to test suspended windows without it.

lib/gzip_vas.c
```diff
@@ -319,28 +327,88 @@ int nxu_run_job(nx_gzip_crb_cpb_t *cmdp, nx_devp_t nxhandle)
...
+		}
+
+		/* On PowerVM, the failures may be caused by a suspended window.
+		   Try a few times and then check the number of credits. If the
+		   system is oversubscribed or the number of total credits
+		   reduced from the time the window was created, then our window
+		   may be suspended. There's no way to be sure, so be
+		   pessimistic and assume it is. */
+		if (nx_config.virtualization == POWERVM &&
+		    ticks_total > (freq >> 5) &&
```
tuliom, line 351: It's a good idea to move `(freq >> 5)` to a macro with a proper name, giving a better idea of what is going on, e.g. `SUSPENDED_TIME_THRESHOLD`.

tuliom (changes requested):

lib/gzip_vas.c
mscastanho: I had actually wrapped these calls on if statements to check if paste_addr and fd are not invalid, to make the function idempotent. But looks like the ifs got lost during a rebase. Fixed on the next push.

lib/gzip_vas.c
mscastanho: Fixed.

test/test_exhaust_credits
mscastanho: Fixed.

test/test_window_suspend.c
mscastanho: Fixed.

test/test_window_suspend.c
mscastanho: Fixed.

lib/gzip_vas.c
```diff
@@ -167,12 +167,18 @@ int nx_function_end(nx_devp_t nxhandle)
...
-		return rc;
-	}
-	close(nxhandle->fd); /* see issue 164 comment */
+		rc = munmap(nxhandle->paste_addr - 0x400, 4096);
+		if (rc < 0) {
+			fprintf(stderr, "munmap() failed, errno %d\n", errno);
+			return rc;
+		}
+
+		rc = close(nxhandle->fd); /* see issue 164 comment */
```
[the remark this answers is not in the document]
mscastanho: Fixed.

lib/gzip_vas.c
mscastanho: Done.

lib/nx_zlib.c
```diff
@@ -562,6 +562,18 @@ nx_devp_t nx_open(int nx_id)
 	if ((nx_id < -1) || (nx_id >= nx_dev_count))
 		nx_id = -1;
 
+	/* On PowerVM we always leave 1 credit available to help handle paste
```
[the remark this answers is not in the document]
mscastanho: @tuliom Fixed. I actually moved this check to the beginning of the function, so it applies for all cases.

mscastanho:
Pushed changes requested so far and a fix for a build error on RHEL7.

test/Makefile.am
```diff
@@ -25,6 +25,14 @@ TESTS = $(test_exe) $(test_scripts) \
 check_PROGRAMS = $(test_exe) $(selector_tests) $(other_tests)
 check_SCRIPTS = $(test_scripts)
 
+# Programs not built/run by default
+EXTRA_PROGRAMS = exhaust_credits exhaust_credits_nx
+EXTRA_SCRIPTS = test_exhaust_credits
+
+exhaust_credits_SOURCES = exhaust_credits.c credit_utils.c test_utils.c
+exhaust_credits_nx_SOURCES = exhaust_credits.c credit_utils.c test_utils.c
+exhaust_credits_nx_CPPFLAGS = -DWITHPREFIX
```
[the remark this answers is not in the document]
tuliom: I interpret this change as having 2 different testsuites now.
Having separate summary LGTM.
Regarding the second issue: does it work if you run `make check TESTS=test_exhaust_credits -C tests/serial-tests` ?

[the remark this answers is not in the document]
mscastanho: Yes, it does 🤦🏼 
Ok, so problems "solved" then. Let me rebase that commit.

mscastanho:
Now both `test_exhaust_credits` and `test_window_suspend` have been integrated to make check. Both will run serially and only on PowerVM. Also `test_window_suspend` is skipped if not run with super-user privileges.

mscastanho:
Fixed build in separate directory and on RHEL7.

test/serial-tests/exhaust_credits.c
```diff
@@ -0,0 +1,149 @@
...
+int main(int argc, char** argv) {
+	int total_credits, used_credits;
+	int ret = 0;
+
+	/* The credit system is only used by PowerVM. So skip otherwise. */
+	if (!is_powervm())
+		return TEST_SKIP;
+
+	if (read_credits(&total_credits, &used_credits))
+		return RET_ERROR;
```
tuliom, line 130: So, when using a kernel that doesn't provide the sysfs interface, this test fails.
LGTM.

test/serial-tests/test_window_suspend.c
```diff
@@ -0,0 +1,186 @@
...
+	int status;
+
+	/* The credit system is only used by PowerVM. So skip otherwise. */
+	if (!is_powervm())
+		return TEST_SKIP;
+
+	/* Need to be root to run drmgr */
+	if (geteuid() != 0) {
+		fprintf(stderr, "This test needs to be run as root.\n");
+		return TEST_SKIP;
```
tuliom, line 103: This is not going to fly....
In most cases this test will be skipped because nobody runs `make check` as root and nobody should.
I understand the problem, but we might need to look at a simple container-like structure to guarantee this test is properly executed.
Considering this won't fail, LGTM for now.

tuliom (approved):

PR #169, opened 2022-05-05
Fix a few issues and add more files to oct

tuliom:
While I was integrating the oct files to snappy, I identified a few issues:

The first 2 patches fix small issues.
The 3rd patch rewrites gzread and make the 4th patch possible.
The 4th patch fix an issue where SW and NX functions were used for the same stream.
5th patch removes functions that became obsolete after the 4th patch.
Finally, the 6th patch add the new files to oct.

### DIFF
--- a/lib/nx_gzlib.c
+++ b/lib/nx_gzlib.c
@@ -34,6 +34,12 @@
  * limitations under the License.
  *
  */
+
+/** @file nx_gzlib.c
+ * \brief Implement the gz* functions for the NX GZIP accelerator and
+ * related functions.
+ */
+
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <unistd.h>
@@ -45,8 +51,6 @@
 #define BUF_LEN 1024
 #define DEF_MEM_LEVEL 8
 #define MAX_LEN 4096
-
-gzFile nx_gzopen_(const char* path, int fd, const char *mode);
 
 bool is_deflate;
 
@@ -61,7 +65,7 @@ typedef struct gz_state {
 	uint8_t * cur_buf;
 } *gz_statep;
 
-int nx_gzclose(gzFile file)
+int __gzclose(gzFile file, int force_nx)
 {
 	gz_statep state = (gz_statep) file;
 	z_streamp stream = &(state->strm);
@@ -78,6 +82,16 @@ int nx_gzclose(gzFile file)
 		return Z_MEM_ERROR;
 
 	if (is_deflate) {
+		typeof(deflate) (*def);
+		typeof(deflateEnd) (*defEnd);
+		if (force_nx) {
+			def = &nx_deflate;
+			defEnd = &nx_deflateEnd;
+		} else {
+			def = &deflate;
+			defEnd = &deflateEnd;
+		}
+
 		/* Finish the stream.  */
 		stream->next_in = Z_NULL;
 		stream->avail_in = 0;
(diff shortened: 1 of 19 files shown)


lib/nx_gzlib.c
```diff
@@ -173,51 +203,80 @@ gzFile nx_gzopen_(const char* path, int fd, const char *mode)
...
 			return 0;
 		}
 	} while (stream->avail_out != 0);
 
-	/* gzread should return when len bytes are wrote on buf, if we still
-	   have unprocessed data on avail_in we need to save it on fifo_in.  */
-	if (stream->avail_in > 0){
-		nx_streamp s = (nx_streamp) stream->state;
-		copy_data_to_fifo_in(s);
-	}
```
mscastanho, line 214: I didn't get why this block isn't needed anymore. Even though the loop may process leftover input bytes from previous inflate calls, the code exits from it if there's no output space left, which can still happen now.

tuliom: @mscastanho That's because we are not freeing/allocating the input buffer between `__gzread()` calls.
If we leave because it needs more output space, all the input data will be kept at `state->cur_buf` and we can just call `inflate()` again the next time `__gzread()` is called.

lib/nx_gzlib.c
```diff
@@ -185,14 +203,33 @@ gzFile nx_gzopen_(const char* path, int fd, const char *mode)
 	return file;
 }
 
-int nx_gzread(gzFile file, void *buf, unsigned len)
+gzFile nx_gzopen_(const char* path, int fd, const char *mode) {
```
rzinsly, line 206: We don't need `nx_gzopen_` anymore, it got replaced by `__gzopen_` in this commit.

lib/nx_gzlib.c
```diff
@@ -185,14 +203,33 @@ gzFile nx_gzopen_(const char* path, int fd, const char *mode)
...
 	z_streamp stream = &(state->strm);
 	int ret, err;
 	const uInt max = len > 10? 10 : 1;
 	uLong last_total_out = stream->total_out;
+	typeof(inflate) (*inf);
 
+	if (force_nx)
+		inf = &nx_inflate;
+	else
+		inf = &inflate;
```
rzinsly, line 232: `__gzclose` has multiple calls to other functions, I think it would be cleaner if you set the function pointers like here.

lib/nx_gzlib.c
```diff
@@ -225,7 +262,21 @@ int nx_gzread(gzFile file, void *buf, unsigned len)
...
 }
 
-int nx_gzwrite (gzFile file, const void *buf, unsigned len)
+int nx_gzread(gzFile file, void *buf, unsigned len) {
+	return __gzread(file, buf, len, 1);
+}
+
+/** \brief Write to a compressed file. Internal implementation.
+ *
+ *  @param file nx_streamp to be processed.
```
rzinsly, line 271: pedantic: file is actually a gzip format file descriptor and the internal state have a stream.

rzinsly (changes requested):

lib/nx_gzlib.c
```diff
@@ -82,37 +92,46 @@ int nx_gzclose(gzFile file)
...
-
-gzFile nx_gzopen(const char* path, const char *mode)
+int nx_gzclose(gzFile file)
 {
-	return nx_gzopen_(path, 0, mode);
+	return __gzclose(file, 1);
 }
 
-gzFile nx_gzopen_(const char* path, int fd, const char *mode)
+gzFile __gzopen_(const char* path, int fd, const char *mode, int force_nx)
```
mscastanho, line 134: Can we get rid of the  trailing `_` and simply use `__gzopen` for the name? That way it's consistent with `__gzwrite` and `__gzclose`.

mscastanho (changes requested):
I agree with @rzinsly's comments. I just have an additional nit.
The rest looks good.

lib/nx_gzlib.c
tuliom: Done.

lib/nx_gzlib.c
tuliom: Modified.

lib/nx_gzlib.c
tuliom: Done.

rzinsly (approved):

mscastanho (approved):

PR #170, opened 2022-05-13
Misc fixes

mscastanho:
Miscellaneous fixes from #168. I separated them into this new PR to facilitate the review process. More info in the commit messages.


inc_nx/nxu.h
```diff
@@ -860,15 +860,6 @@ typedef struct {
...
 
-/* prototypes */
-#ifdef NX_JOB_CALLBACK
-int nxu_run_job(nx_gzip_crb_cpb_t *c, void *handle,
-		int (*callback)(const void *));
-#else
-int nxu_run_job(nx_gzip_crb_cpb_t *c, void *handle);
-#endif
-
-
```
tuliom, line 871: File `samples/gzip_nxfht.c` needs to be modified too after this removal.

tuliom (changes requested):
Looks mostly good to me, except for a minor detail missing in the last patch.

inc_nx/nxu.h
mscastanho: @tuliom Done. I also fixed `samples/gzip_nxdht.c` and `samples/gunzip_nx_test.c`. All of them pull some parts from libnxz.a and also reimplement other parts. I tried to avoid increasing this dependency on libnxz.a but also didn't improve the situation much further. At least they will still compile and run as before.

In the future we should find a way to make these completely standalone to avoid such conflicts. But this is work for a separate PR.

doc/Addendum-NX-GZIP-for-PowerVM.txt
```diff
@@ -220,14 +220,14 @@ struct vas_tx_win_open_attr {
...
   Adding the following sysfs interfaces to the user-space:
 
-/sys/kernel/vas/VasCaps/VDefGzip:    (Called default GZIP capabilities)
-	avail_lpar_creds  target_lpar_creds  used_lpar_creds
+/sys/devices/virtual/misc/vas/vas0/gzip/default_capabilities/    (Called default GZIP capabilities)
+	nr_total_credits nr_used_credits
 
   On our development system with 16 cores and dedicated: these entry
-  values are 320 320 0 respectively
+  values are 320 320 respectively
```
rzinsly, line 227: Does that mean nr_used_credits starts at 320? Is that correct?

lib/nx_zlib.h
```diff
@@ -181,15 +181,12 @@ struct nx_dev_t {
...
-		void *paste_addr;
-		int fd;
-		void *vas_handle;
-	}; /* vas */
+	int16_t version;
+	int16_t id;
+	int64_t flags;
+	void *paste_addr;
+	int fd;
+	int function;
```
rzinsly, line 189: Would be good to have a comment in this block indicating that this is the vas handle.

samples/gunzip_nx.c
```diff
@@ -58,8 +58,9 @@
 #include <errno.h>
 #include <signal.h>
 #include "nxu.h"
-#include "nx.h"
+/* #include "nx.h" */
```
rzinsly, line 61: This comment could be removed from all sample files.

rzinsly (changes requested):
I believe we could remove nx.h as it is not used anywhere anymore.
Besides that I have a few comments.

doc/Addendum-NX-GZIP-for-PowerVM.txt
mscastanho: Nice catch. No, it's not correct. I should've deleted the second `320` instead. I'll fix it.

lib/nx_zlib.h
mscastanho: But the "vas handle" does not exist anymore. This struct is our handle to communicate with NX, and past_addr, fd and function are just parts of that NX handle. I don't see how keeping the distinction is useful.

doc/Addendum-NX-GZIP-for-PowerVM.txt
mscastanho: Fixed.

samples/gunzip_nx.c
mscastanho: Done.

mscastanho:
> I believe we could remove nx.h as it is not used anywhere anymore.
@rzinsly Done.

lib/nx_zlib.h
rzinsly: I think it would be useful for new developers. The `version`, `id`, and `flags` fields are vas specific attributes to be used on ioctl, and `past_addr` is the vas memory address.
`fd` is actually the device file descriptor so that and `function` are not directly part of vas.
Maybe not keeping a distinction between what is vas related and the rest but at least adding comments describing the fields should help.

rzinsly (approved):

lib/nx_zlib.h
mscastanho: @rzinsly Done. I added comments for each field and ended up removing `version`, `id` and `flags`. `open_device_nodes` just uses hardcoded values for `version` and `flags` when doing the ioctl, and `id` is given as parameter to that function, so the fields in this struct are not being used.

rzinsly: @mscastanho Perfect!
LGTM, thanks!

tuliom (approved):

PR #172, opened 2022-05-31
Some samples/ cleaning

rzinsly:
In order to stop using internal functions I had to create smaller versions of some of these functions, I'm open to suggestions to fully remove them.

### DIFF
--- a/samples/Makefile.am
+++ b/samples/Makefile.am
@@ -1,7 +1,5 @@
 AM_CPPFLAGS = -I$(top_srcdir)/inc_nx -I$(top_srcdir)/lib
-LDFLAGS = -static-libtool-libs
-
-STATIC_LIBNXZ = ../lib/libnxz.la
+LDADD= ../lib/libnxz.la $(PTHREAD_LIBS) -ldl
 
 EXTRATESTS = gunzip_nx_test gzip_nxfht_test gzip_nxdht_test compdecomp_th \
 	bad_irq_check rand_pfault_check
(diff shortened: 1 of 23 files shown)


samples/samples_utils.c
```diff
@@ -0,0 +1,215 @@
+#include <stdio.h>
+#include <string.h>
+#include <assert.h>
+#include <sys/platform/ppc.h>
+#include "nxu.h"
+#include "nx_dht.h"
+#include "nx_dbg.h"
+#include "nx_zlib.h"
```
tuliom, line 8: Do we really need these internal headers?

samples/Makefile.am
```diff
@@ -1,7 +1,5 @@
 AM_CPPFLAGS = -I$(top_srcdir)/inc_nx -I$(top_srcdir)/lib
-LDFLAGS = -static-libtool-libs
-
-STATIC_LIBNXZ = ../lib/libnxz.la
+LDADD= ../lib/libnxz.la $(PTHREAD_LIBS) -ldl
```
mscastanho, line 2: If I undestand correctly, these samples are in this repo just as examples of how to interact directly with NX Gzip (mostly interesting to us, libnxz devs), and they replicate a lot of libnxz's internal code.  So does it make sense to keep linking them against libnxz here?

mscastanho (commented):
I agree this is an improvement compared to what we have today. But ultimately, I'd prefer if these samples lived in a separate repo and didn't rely on libnxz at all. Having them here may confuse users and give the false impression that they are samples of how to use libnxz, which they are not.

Note I'm not asking you to do this yourself, I just wanted to raise this topic so we can clarify what's the best thing to do regarding these sample programs.

samples/Makefile.am
tuliom: > mostly interesting to us, libnxz devs
I disagree with this statement.
I believe there is a rare kind of user that may be interested in manipulating the NX GZIP more directly.  If that was not right, it wouldn't make sense to export the `nx_*` ABI in libnxz, right?

rzinsly: I think there is still value to have a directory with examples on how to use the NX so people interested in it don´t need to dig into libnxz's code. The problem is that currently the `samples` need a lot more cleaning, probably having just few files like zlib's `examples` would be better.
We also need a `README` file to explain the purpose of the directory and avoid the impression that it is about libnxz's usage.

samples/samples_utils.c
rzinsly: There is a lot of `nx_dht.h` structures used by samples, and the nx state from `nx_zlib.h` is necessary to get access to the device handler.
I removed `nxu.h` and `nx_dbg.h` from this file.

samples/Makefile.am
tuliom: > One may want more control over the engine (compared to the zlib compat API) without going into all this trouble, that's why the nx_* functions exist.

@mscastanho I completely agree with this part and that's why I think sample files showing how to do that is very important. We don't have that yet, but I think this PR is in the right direction.
Although I wouldn't be surprised if we had an user interested in having far more control of how to manipulate the NX device.

samples/samples_utils.c
tuliom: OK. What if we add a comment here with a TODO saying: "Stop using the following nx headers"?

samples/samples_utils.h
```diff
@@ -0,0 +1,20 @@
+#include <stdio.h>
+#include <string.h>
+#include <assert.h>
+
+typedef struct top_sym_t {
```
tuliom, line 5: Likewise, should we add a source comment saying this struct should not be used and adding a TODO to remove it?

tuliom (approved):

mscastanho (approved):

PR #176, opened 2022-06-07
Dynamically define number of threads for test

mscastanho:
If the TEST_NTHREADS envvar is not provided, try to choose a more
appropriate value depending on the virtualization type. On baremetal we
use a larger number of threads, but on PowerVM we try to limit to the
number of credits available, or a lower number if the the credit
information is not available on sysfs.


test/test_multithread_stress.c
```diff
@@ -187,36 +165,62 @@ static int run_case(void* _pstats)
...
+			    &total_credits) ||
+	   read_sysfs_entry(SYSFS_VAS_CAPS "nr_used_credits",
+			    &used_credits))
+		/* Failed to read from sysfs, so fallback to number of
+		   processors */
+		return get_nprocs();
+
+	if (total_credits != used_credits)
+		/* Use as many threads as available credits */
+		return total_credits - used_credits;
```
tuliom, line 186: I didn't understand this part. This test is expected to consume only 1 credit because threads will reuse the VAS window.
So, why should this test use as many threads as available credits?
Should the test be executed with a lower value for `max_vas_reuse_count`?
I don't think the original code was meant to test credit exhaustion.  Just a kind of stress test.

Or maybe the number of credits available should only indicate a maximum value for `get_nthreads()`, e.g. in a system with 1000 threads, the chances of consuming 2 or more credits are too high, so this test should fail if there are not enough credits available.

And if the the last interpretation is correct, should we move this test to the `serial-tests`?

mscastanho: @tuliom Fair ponit. My mind was still set to think in terms of 1 credit per thread/process because of the DLPAR work. But I agree it doesn't apply here.

The main purpose of this test is to check if several threads can run simultaneously and succeed even in the face of contention, right? And the idea behind this initial code to choose the number of threads is just to make sure the test is able to open VAS windows for all threads.

> Should the test be executed with a lower value for max_vas_reuse_count?

I don't think this brings much more value to the test. We want to force contention, we can do that using a single window.

So let's take one step back: the original problem we had with this test on PowerVM was that it tried to create more threads than credits available, _because window reuse was not enabled_. Now we do window reuse by default, so since our current `max_vas_reuse_count` is 10000, on PowerVM we could in theory run 10k threads consuming only one credit, which seems more than enough to guarantee the test succeeds even on systems with high processor counts.

So is there actually anything to be done here? If we really want to be on the safe side, we could calculate nprocs / max_vas_reuse_count and check if there are that many credits available in the system.

tuliom: > Now we do window reuse by default, so since our current `max_vas_reuse_count` is 10000, on PowerVM we could in theory run 10k threads consuming only one credit, which seems more than enough to guarantee the test succeeds even on systems with high processor counts.

@mscastanho I don't agree with this. In the default settings, each thread iterates 6x, compressing and decompressing.
That's *at least* 12 NX jobs / thread, giving a maximum of 10000 / 12 = 833 threads.
Large systems have more than 900 threads.

mscastanho: @tuliom my reasoning was flawed. The window will be reallocated when it has been reused `max_vas_reuse_count`. I changed the code to calculate rough amount of windows will be used to estimate credit usage. It should only make a difference for servers with a huge amount of threads.

tuliom (approved):

PR #180, opened 2022-06-17
Correctly handle suspension of shared windows

mscastanho:
This PR is intended to fix the following issue:
https://github.com/libnxz/power-gzip/blob/3d08904a072d76f80a6229cb94847a43fc3fe58a/lib/gzip_vas.c#L371-L372

When libnxz suspects a VAS window might be suspended, it tries to open a
new one. But since the kernel does not allow the creation of new windows
while there are still suspended ones, libnxz has to close its existing
window before opening the new one.

This causes problems when a window is shared by multiple threads,
because one thread may issue a paste request on the window's associated
memory address, but it may not be valid anymore because a parallel
thread has already munmap'd it.

To avoid this problem every window now has a read-write lock and threads
are required to lock it first before performing any action on the
window, like reopening it or issuing a paste.

I ran `compdecomp_th` on PowerVM (100 threads) and baremetal (1000 threads) and couldn't detect any significant performance penalty.

### DIFF
--- a/lib/gzip_vas.c
+++ b/lib/gzip_vas.c
@@ -298,16 +298,25 @@ int nxu_run_job(nx_gzip_crb_cpb_t *cmdp, nx_devp_t nxhandle)
 	int ret=0, retries=0, wait_count=0;
 	uint64_t ticks_total = 0, wait_ticks = 0;
 	uint64_t freq = nx_get_freq();
-	int used_credits, total_credits;
+	int used_credits, total_credits, window_generation=-1;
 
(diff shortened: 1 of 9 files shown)


tuliom (approved):
The changes LGTM, but it appears to me that commits 2 and 3 are in the wrong order, i.e. if the test from commit 2 causes libnxz to fail until commit 3 is applied, then the fix should be applied before the test in order to avoid `git-bised` failures.

tuliom (approved):

rzinsly (approved):

Issue #181, opened 2022-06-22
git error using libnxz


brihh:
are there any installed tools that can be used to demonstrate libnxz? git uses libz, so i tried it, but i get errors:

[root@p182n11 aa]# export LD_PRELOAD=/usr/lib64/libnxz.so
[root@p182n11 aa]# ldd `type -p git`;time git clone https://github.com/libnxz/power-gzip
        linux-vdso64.so.1 (0x00007fff8cf80000)
        /usr/lib64/libnxz.so (0x00007fff8ca20000)
        libpcre2-8.so.0 => /lib64/libpcre2-8.so.0 (0x00007fff8c960000)
        libz.so.1 => /lib64/libz.so.1 (0x00007fff8c920000)
        libpthread.so.0 => /lib64/glibc-hwcaps/power9/libpthread-2.28.so (0x00007fff8c8d0000)
        librt.so.1 => /lib64/glibc-hwcaps/power9/librt-2.28.so (0x00007fff8c8a0000)
        libcrypto.so.1.1 => /lib64/libcrypto.so.1.1 (0x00007fff8c520000)
        libc.so.6 => /lib64/glibc-hwcaps/power9/libc-2.28.so (0x00007fff8c310000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007fff8c2e0000)
        /lib64/ld64.so.2 (0x00007fff8cfa0000)
Cloning into 'power-gzip'...
remote: Enumerating objects: 3908, done.
remote: Counting objects: 100% (909/909), done.
remote: Compressing objects: 100% (335/335), done.
BUG: zlib.c:57: total_in mismatch
fatal: fetch-pack: invalid index-pack output

real    0m3.941s
user    0m0.011s
sys     0m0.000s

[root@p182n11 aa]# unset LD_PRELOAD
[root@p182n11 aa]# cd /lib64
[root@p182n11 lib64]# ll libz*
lrwxrwxrwx. 1 root root      19 Sep 24  2021 libzhuyin.so.13 -> libzhuyin.so.13.0.0
-rwxr-xr-x. 1 root root 1266536 Sep 24  2021 libzhuyin.so.13.0.0
lrwxrwxrwx. 1 root root      14 Apr 12 04:48 libz.so -> libz.so.1.2.11
lrwxrwxrwx. 1 root root      14 Apr 12 04:48 libz.so.1 -> libz.so.1.2.11
[...]

PR #182, opened 2022-06-22
Stop to subtract from total_in in nx_inflate()

tuliom:
Partially fix issue #181 by stopping to subtract from total_in in
nx_inflate().  total_in and total_out should always increase or be left
unchanged during deflate/inflate calls.

Also add tests to validate the values of:
 - avail_in
 - avail_out
 - next_in
 - next_out
 - total_in
 - total_out

### DIFF
--- a/lib/nx_deflate.c
+++ b/lib/nx_deflate.c
@@ -38,7 +38,6 @@
 
 #include <stdio.h>
 #include <stdlib.h>
-#include <malloc.h>
 #include <string.h>
 #include <unistd.h>
 #include <stdint.h>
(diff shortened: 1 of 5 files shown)


rzinsly (approved):

mscastanho (approved):

tuliom:
@rzinsly , @mscastanho I updated the PR with more tests in 2 deflate calls that I had missed as well as being more pedantic about the value in `avail_in` and `avail_out`.
I also added a commit that removes unnecessary inclusion of `malloc.h`.

rzinsly (approved):

mscastanho (approved):

Issue #187, opened 2022-07-15
zlib may erroneously call libnxz functions


mscastanho:
Recently I hit the following scenario: libnxz's `deflateInit2_` calls zlib's init function `sw_deflateInit2` (which maps to `deflateInit2_` from libz.so), which internally calls `deflateReset`. Since libnxz has been loaded first, this maps to libnxz's `deflateReset`, instead of zlib's.

```
$ cat ./bug.c 
#include <assert.h>
#include <stdio.h>
#include <stdlib.h>
#include <zlib.h>

int main()
{
        Byte src[128], dst[128];
        z_stream strm;

        strm.zalloc = Z_NULL;
        strm.zfree = Z_NULL;
        strm.opaque = (voidpf)0;
        assert(deflateInit(&strm, Z_DEFAULT_COMPRESSION) == Z_OK);

        return 0;
}

$ gcc -Og -g -o bug bug.c -lz
$ LD_PRELOAD=./lib/.libs/libnxz.so gdb ./bug
> break deflateReset
> run

Call stack:
[0] from 0x00007ffff7eee3ac in deflateReset+8 at nx_deflate.c:2312   <-------- called libnxz (not good!)
[1] from 0x00007ffff7e69c58 in deflateInit2_+632 at deflate.c:362    <-------- was on zlib
[2] from 0x00007ffff7f05fd8 in sw_deflateInit2_+48 at sw_zlib.c:102
[3] from 0x00007ffff7eedfd8 in deflateInit2_+276 at nx_deflate.c:2269
[4] from 0x00007ffff7eee384 in deflateInit_+48 at nx_deflate.c:2251
[5] from 0x00000000100102a8 in main+60 at bug.c:14
```

I tried passing `RTLD_DEEPBIND` to dlopen to force the dlopen'd zlib to only call it's own symbols:
```diff
diff --git a/lib/sw_zlib.c b/lib/sw_zlib.c
index e97b477..09c0173 100644
--- a/lib/sw_zlib.c
+++ b/lib/sw_zlib.c
@@ -285,7 +285,7 @@ int sw_zlib_init(void)
 {
        char *error;
 
-       sw_handler = dlopen(ZLIB_PATH, RTLD_LAZY);
+       sw_handler = dlopen(ZLIB_PATH, RTLD_LAZY | RTLD_DEEPBIND);
        if(sw_handler == NULL) {
                prt_err(" %s\n", dlerror());
                return Z_ERRNO;
```

[...]

PR #188, opened 2022-07-18
Make *Reset functions always restore NX state

mscastanho:
Please read commit messages for more details.

This is marked as draft because I haven't found a solution for #187 yet. This PR is affected by that issue in two situations:
1. libnxz calls `sw_deflateInit2_`, then zlib calls `deflateReset`, which ends up calling libnxz's `deflateReset`.
2. On `deflateReset` libnxz calls `sw_deflateReset`, which calls `deflateResetKeep`, which ends up using libnxz's `deflateResetKeep`.

The first one is easier to workaround because there won't be an entry in `stream_map`, so we now we have to call zlib's `deflateReset`. I couldn't find a way to detect calls from zlib for case 2 yet.

### DIFF
--- a/lib/nx_inflate.c
+++ b/lib/nx_inflate.c
@@ -59,6 +59,7 @@
 #include "nx-gzip.h"
 #include "nx_zlib.h"
 #include "nx_dbg.h"
+#include "nx_map.h"
 
 /** \brief Fixed 32K history length
  * \details The maximum distance in the deflate standard is 32768 Bytes.
@@ -277,7 +278,6 @@ int nx_inflate(z_streamp strm, int flush)
 {
 	int rc = Z_OK, in, out;
 	nx_streamp s;
-	void *temp = NULL;
 
 	if (strm == Z_NULL) return Z_STREAM_ERROR;
 	s = (nx_streamp) strm->state;
@@ -286,28 +286,15 @@ int nx_inflate(z_streamp strm, int flush)
 	/* check for sw deflate first*/
 	if (has_nx_state(strm) && s->switchable
 	    && (0 == use_nx_inflate(strm, flush))) {
-		/*Use software zlib, switch the sw and hw state*/
+		/* Use software zlib, switch the sw and hw state */
 		s = (nx_streamp) strm->state;
 		s->switchable = 0; /* decided to use sw zlib and not switchable */
-		temp  = s->sw_stream;  /* save the sw pointer */
-		s->sw_stream = NULL;
+		strm->state = s->sw_stream;  /* use sw internal state */
 
-		rc = nx_inflateEnd(strm); /* free the hw resource */
-		prt_info("call nx_inflateEnd to clean the hw resource,rc=%d\n",rc);
-		strm->state = temp;  /* restore the sw pointer */
 		prt_info("call software inflate,len=%d\n", strm->avail_in);
 		rc = sw_inflate(strm,flush);
 		prt_info("call software inflate, rc=%d\n", rc);
 		return rc;
-	}else if(s->sw_stream){
-		/*decide to use nx here, release the sw resource */
-		temp  = (void *)strm->state;
-		strm->state = s->sw_stream;
-
-		rc = sw_inflateEnd(strm);
-		prt_info("call sw_inflateEnd to clean the sw resource,rc=%d\n",rc);
-		strm->state = temp;
-		s->sw_stream = NULL;
 	}
 
 	s->switchable = 0;
@@ -2000,7 +1987,7 @@ int inflateInit_(z_streamp strm, const char *version, int stream_size)
 int inflateInit2_(z_streamp strm, int windowBits, const char *version, int stream_size)
 {
 	int rc;
-	void *temp = NULL;
+	void *sw_state = NULL;
 	nx_streamp s;
 
 	/* statistic */
@@ -2013,31 +2000,50 @@ int inflateInit2_(z_streamp strm, int windowBits, const char *version, int strea
 		if (rc != Z_OK)
 			return rc;
 
-		/*If the stream has been initialized by sw*/
-		if (strm->state && (0 == has_nx_state(strm))) {
-			temp = (void *)strm->state; /*record the sw context*/
-			strm->state = NULL;
-			prt_info("this stream has been initialized by sw\n");
-		}
+		sw_state = (void *)strm->state; /* keep sw context pointer */
+		strm->state = NULL;
+		prt_info("sw state initialized and ready for fallback\n");
+
+		/* Save mapping between z_streamp and internal states */
+		struct stream_map_entry *sme =
+			malloc(sizeof(struct stream_map_entry));
+		if (sme == NULL)
+			return Z_MEM_ERROR;
+
+		/* Save stream parameters in case we need to restart NX
+		   on inflateReset */
+		sme->windowBits = windowBits;
+		strcpy(sme->version, version);
+		sme->stream_size = stream_size;
 
 		rc = nx_inflateInit2_(strm, windowBits, version, stream_size);
-		if (rc != Z_OK) {
-			strm->state = temp;
+		if (rc == Z_OK) {
+			s = (nx_streamp) strm->state;
+			s->switchable = 1;
+			s->sw_stream = sme->sw_state = sw_state;
+			sme->hw_state = s;
+		} else {
+			strm->state = sw_state;
 			if (rc == Z_STREAM_ERROR && errno == EAGAIN) {
 				/* Failed due to lack of credits on POWERVM */
 				prt_info("Falling back to software compression\n");
+				sme->hw_state = NULL;
+				sme->sw_state = sw_state;
 				/* Since internal state is already set to
 				   zlib's, there's nothing left to do. */
 				rc = Z_OK;
 			} else {
 				/* release the sw stream */
 				(void) sw_inflateEnd(strm);
+				free(sme);
+				return rc;
 			}
-		} else if (temp) { /* recorded sw context*/
-			s = (nx_streamp) strm->state;
-			s->sw_stream = temp;
-			s->switchable = 1;
 		}
+
+		/* This should only fail if there's not enough memory to resize
+		   the table */
+		if (nx_map_put(stream_map, strm, sme))
+			return Z_MEM_ERROR;
 	} else if (nx_config.mode.inflate == GZIP_NX) {
 		rc = nx_inflateInit2_(strm, windowBits, version, stream_size);
 	} else {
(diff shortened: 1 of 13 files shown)


mscastanho:
I believe the PR is now ready for review. Putting zlib in a different
link namespace fixed issue #187, so I included it here. I also added
tests for the reset functions to verify NX state is always being
reset.

Note to reviewers: commit `Make *Reset functions always restore NX state`
seems big, but the changes to nx_inflate.c and nx_deflate.c are pretty
much the same.

lib/nx_inflate.c
```diff
@@ -2046,58 +2052,160 @@ int inflateInit2_(z_streamp strm, int windowBits, const char *version, int strea
...
+				 int (*hw_reset)(z_streamp))
 {
 	int rc;
 
-	if (0 == has_nx_state(strm)){
-		rc = sw_inflateReset(strm);
-	}else{
-		rc = nx_inflateReset(strm);
-	}
+	if (nx_config.mode.deflate == GZIP_AUTO) {
```
rzinsly, line 2061: Should be nx_config.mode.**inflate**

lib/nx_inflate.c
```diff
@@ -2046,58 +2052,160 @@ int inflateInit2_(z_streamp strm, int windowBits, const char *version, int strea
...
+
+int inflateReset2(z_streamp strm, int windowBits)
 {
 	int rc;
 
-	if (0 == has_nx_state(strm)){
-		rc = sw_inflateResetKeep(strm);
-	}else{
-		rc = nx_inflateResetKeep(strm);
+	if (nx_config.mode.deflate == GZIP_AUTO) {
```
rzinsly, line 2124: Likewise

lib/nx_inflate.c
```diff
@@ -2046,58 +2052,160 @@ int inflateInit2_(z_streamp strm, int windowBits, const char *version, int strea
...
-
 int inflateEnd(z_streamp strm)
 {
 	int rc;
 
 	/* statistic */
 	zlib_stats_inc(&zlib_stats.inflateEnd);
 
-	if (0 == has_nx_state(strm)){
+	if (nx_config.mode.deflate == GZIP_AUTO) {
```
rzinsly, line 2179: Likewise.

lib/nx_inflate.c
```diff
@@ -2046,58 +2052,160 @@ int inflateInit2_(z_streamp strm, int windowBits, const char *version, int strea
...
+		struct stream_map_entry *sme = nx_map_remove(stream_map, strm);
+		validate_stream_map_entry(sme, strm);
+
+		/* TODO: What to do about the delay? */
+		/* if (0 != avg_delay) */
+		/* 	decrease_delay(); */
```
rzinsly, line 2185: Why not decrease it when there is no nx state?

rzinsly (changes requested):
On commit 0e90890 the last paragraph has a confuse line:
"This way libnxz we can call zlib symbols directly through the handle..."
"This way we can" or "This way libnxz can" reads better.

And I wonder if we need the map or it would work well just keeping the last entry.

lib/nx_inflate.c
```diff
@@ -2046,58 +2052,160 @@ int inflateInit2_(z_streamp strm, int windowBits, const char *version, int strea
...
+		strm->state = sme->sw_state;
+		rc = sw_inflateEnd(strm);
+		prt_info("call sw_inflateEnd,rc=%d\n", rc);
+
+		strm->state = (void *) sme->hw_state;
+
+		/* Free NX state */
+		if (sme->hw_state) {
+			sme->hw_state->sw_stream = NULL;
+			/* TODO: what to return if sw_deflateEnd failed? */
```
rzinsly, line 2197: sw_**inflate**End

lib/nx_inflate.c
mscastanho: Agreed. I did this on nx_deflate.c but forgot to do the same here. Thanks!

lib/nx_inflate.c
mscastanho: Fixed.

lib/nx_inflate.c
mscastanho: Fixed.

mscastanho:
> And I wonder if we need the map or it would work well just keeping the last entry.

@rzinsly I'm afraid we need the map for two reasons:
1. If we create a single global var to save the last entry, all threads in a binary will point to the same field, so they could overwrite each other.
2. Maybe we could save this global variable on TLS, but what if the user wants to maintain more than one stream at once in the same thread? Like for deflating and inflating different streams "at the same time" as it gradually receives data. In that case these multiple streams would all overwrite the same variable. Even though I haven't seen such an example yet, I wanted to avoid excluding that use case.

rzinsly (approved):

mscastanho:
> Turns out test_stress.mix and some other tests are failing when libnxz is linked against older glibcs. It works fine with glibc 2.34 and later, but fails with glibc 2.32 (AT 14) and 2.31 (Debian 11). Bisect points to the commit "fixing" https://github.com/libnxz/power-gzip/issues/187.

@rzinsly @tuliom I investigated further and this is what I found out: `test_stress.sw` and any other test that calls `sw_deflateInit2_` in a multithreaded scenario is hitting a segfault on glibc's `__get_nprocs`. I bisected glibc and the commit that fixes it is:
```
commit b5eeca8cfd9d0fd92b5633a88901d9ff27f2b496
Author: Andreas Schwab <schwab@linux-m68k.org>
Date:   Tue Dec 8 19:17:41 2020 +0100

    Fix parsing of /sys/devices/system/cpu/online (bug 25859)
    
    The file contains comma-separated ranges, not spaces.
```

even though none of the hosts I tested the code on have commas and/or spaces in that file. This commit was included in glibc 2.33. The only distro with an older glibc that still passes the tests is the one from SLES15. Using Advance Toolchain 15 is enough to fix the issue as well. See below:

```
SLES15 (glibc 2.31): OK
RHEL8 (glibc 2.28): FAIL 
RHEL9 (glibc 2.34): OK
Ubuntu 22.04 (glibc 2.35): OK
Ubuntu 20.04 (glibc 2.31): FAIL
Debian 10 (glibc 2.28): FAIL
Debian 11 (glibc 2.31): FAIL
Advance Toolchain 14.0 (glibc 2.32): FAIL
Advance Toolchain 15.0 (glibc 2.34): OK
```
If I run the same test in containers (tested with RHEL8 and Debian 11) it works 🤷🏼 .
Also note this only happens when glibc is in a separate link namespace (`dlmopen(LM_ID_NEWLM,...`).
So I don't think we can ignore this issue, but at the same time I couldn't find a solution for this, or an alternative to solve #187. Ideas?

lib/sw_zlib.c
```diff
@@ -36,7 +36,7 @@
  *
  */
 
-
+#define _GNU_SOURCE
```
tuliom, line 39: This means libnxz won't be portable to other standard C libraries.
LGTM.

lib/nx_map.c
```diff
@@ -0,0 +1,191 @@
+#include <stdlib.h>
```
tuliom, line 1: I expected to see a copyright header here mentioning the 2 licenses of this file/project.

lib/nx_map.c
```diff
@@ -0,0 +1,191 @@
...
+
+	m = malloc(sizeof(nx_map_t));
+	if (!m) {
+		prt_err("Failed to allocate map.\n");
+		return NULL;
+	}
+
+	m->array = calloc(MAP_INIT_SIZE, sizeof(map_entry_t));
+	if (!m->array) {
+		prt_err("Failed to allocate internal array for map.\n");
```
tuliom, line 52: Should `m` be freed here?

tuliom (changes requested):
LGTM in overall, but I noticed a few minor issues.

lib/nx_map.c
mscastanho: Good catch. Fixed.

tuliom (approved):
The changes LGTM.
I still don't know what to do regarding the glibc issue.
I guess we need to collect more details, specially on RHEL 8. IMHO, it's surprising this issue is not fixed there.

tuliom:
I have just updated this PR with a slightly modified implementation.
In this new implementation, I give a different interpretation on how `deflateResetKeep()` and `inflateResetKeep()` should behave.
In his original proposal, @mscastanho interpreted that both these function should behave as their counterparts `deflateReset()` and `inflateReset()`.
While I agree with his implementation and I believe that's the most logical too, it is not practical to move forward with it because of bug #187 .
In this new interpretation `deflateResetKeep()` and `inflateResetKeep()` act as simpler reset functions, i.e. they reset only the stream being used without modifying the other one or recovering the NX stream after switching to software.
This is important in order to avoid that zlib's `deflate()`/`inflate()` ends up calling libnxz's `deflateResetKeep()`/`inflateResetKeep()` which would recover the NX stream while zlib still expects its own stream there.

Note: I noticed a segmentation fault on RHEL 7, but I've been able to reproduce this issue with code already in `develop`.

rzinsly (approved):
The `*Reset_skeleton` functions are not needed anymore but I'm fine with keeping them.
LGTM

tuliom:
> The `*Reset_skeleton` functions are not needed anymore but I'm fine with keeping them. LGTM

Let me remove them.  I'm also planning to improve documentation.
I think this will be important in order to explain the current decisions in this proposal.

tuliom:
Done!

configure
```diff
@@ -16379,7 +17194,7 @@ n|N|no|No|NO)
 ;;
 esac
 
-else
+else $as_nop
```
tuliom, line 17197: I suspect this is changing the behavior of the code.
I have seen cases where some doxygen features were enabled even if doxygen was not available.
While debugging this, I found out that `DX_FLAG_doc` was set and I suspect this is the cause.

The problem is: this is automatically generated.  Should I remove this patch and use older versions of GNU autoconf and GNU automake?

rzinsly: >Should I remove this patch and use older versions of GNU autoconf and GNU automake?

IMO yes, hopefully the next version of autoconf fixes that we can update the version then.

tuliom: Done. I replaced the original patch with a new one that changes just a few lines.
I can't reproduce the issue anymore.

mscastanho:
@tuliom The changes look good, I just didn't understand the reason for this one:
```diff
diff --git a/test/test_reset.c b/test/test_reset.c
index 4625f03..57069d0 100644
--- a/test/test_reset.c
+++ b/test/test_reset.c
@@ -25,16 +25,18 @@
 
 static void validate_stream(z_stream *strm, struct internal_state *sw_state,
                            nx_streamp hw_state) {
+#ifndef KEEP
        /* NX state should always be loaded */
        assert(has_nx_state(strm));
 
-       /* Pointers to internal structures shouldn't change */
-       nx_streamp s = (nx_streamp) strm->state;
-       assert(s == hw_state);
-       assert(s->sw_stream == sw_state);
-
        /* Should be ready for fallback */
-       assert(s->switchable == 1);
+       assert(hw_state->switchable == 1);
+#endif
+
+       if (hw_state->sw_stream) {
+               /* Pointers to internal structures shouldn't change */
+               assert(hw_state->sw_stream == sw_state);
+       }
 }
 
 int main()
```
The way it's currently written we're not checking that the currently loaded NX state is switchable, only the initial one (from before deflateReset), and we're also not checking that it didn't change.

tuliom:
@mscastanho You're right. I shouldn't have changed the code path for `deflateReset()` and `inflateReset()`.  Just for `*ResetKeep()`.

While fixing this, I figured out the code setting `hw_state` and `sw_state` could be set closer to `deflateInit()` and `inflateInit()`.
I also modified the code that zeroes `strm` in order to zero all the structure, instead of a few fields.

These are the new differences:
```
$ $ git diff  7ac8ea2 -- test/test_reset.c
diff --git a/test/test_reset.c b/test/test_reset.c
index 4625f03..e659696 100644
--- a/test/test_reset.c
+++ b/test/test_reset.c
@@ -25,6 +25,12 @@
 
 static void validate_stream(z_stream *strm, struct internal_state *sw_state,
                            nx_streamp hw_state) {
+#ifdef KEEP
+       if (hw_state->sw_stream) {
+               /* Pointers to internal structures shouldn't change */
+               assert(hw_state->sw_stream == sw_state);
+       }
+#else
        /* NX state should always be loaded */
        assert(has_nx_state(strm));
 
@@ -35,6 +41,7 @@ static void validate_stream(z_stream *strm, struct internal_state *sw_state,
 
        /* Should be ready for fallback */
        assert(s->switchable == 1);
+#endif
 }
 
 int main()
@@ -55,9 +62,9 @@ int main()
                return TEST_ERROR;
        }
 
-       strm.zalloc = Z_NULL;
-       strm.zfree = Z_NULL;
-       strm.opaque = (voidpf)0;
+       /* Ensure there is no garbage in the structure, which could cause
+          segmentation fault.  */
+       memset(&strm, 0, sizeof(z_stream));
 
        /* Reseting an uninitialized stream must fail */
        zcheck(CALL_DEFLATE_RESET(&strm), Z_STREAM_ERROR);
@@ -77,11 +84,12 @@ int main()
                        return TEST_SKIP;
                }
 
+               nx_streamp hw_state = (nx_streamp) strm.state;
+               struct internal_state *sw_state = hw_state->sw_stream;
+
                /* Reseting between init and deflate should be OK */
                zcheck(CALL_DEFLATE_RESET(&strm), Z_OK);
 
-               nx_streamp hw_state = (nx_streamp) strm.state;
-               struct internal_state *sw_state = hw_state->sw_stream;
                strm.next_in   = src;
                strm.next_out  = compr;
                strm.avail_in  = rounds[i] == SW ? 100 : src_len;
@@ -102,7 +110,10 @@ int main()
                        return TEST_SKIP;
                }
 
-               /* Reseting between init and deflate should be OK */
+               hw_state = (nx_streamp) strm.state;
+               sw_state = hw_state->sw_stream;
+
+               /* Reseting between init and inflate should be OK */
                zcheck(CALL_INFLATE_RESET(&strm), Z_OK);
 
                strm.next_in   = compr;
```

mscastanho:
@tuliom @rzinsly Turns out I can't review my own PR 😂 
The changes LGTM. So if it's OK with you, I think we can merge this.

rzinsly (approved):

PR #193, opened 2022-08-15
Miscelaneous fixes

tuliom:
- Properly initialize variables.
- Abort `test_multithread_stress` if a thread returns failure.
- Cosmetic improvements.
- Kill all the threads from `test_multithread_stress` after waiting for too long.
- Avoid closing an invalid file descriptor.

### DIFF
--- a/test/test_multithread_stress.c
+++ b/test/test_multithread_stress.c
@@ -229,6 +229,7 @@ int main(int argc, char **argv)
 	printf("Test Iterations:\t%d\n",test_iterations);
 
 	thread_info = (struct stats *) malloc(thread_num*sizeof(struct stats));
+	memset(thread_info, 0, thread_num * sizeof(struct stats));
 
 	if (generate_data_buffer(data_buf) != 0 || thread_info == NULL) {
 		free_data_buffer(data_buf);
--- a/test/test_multithread_stress.c
+++ b/test/test_multithread_stress.c
@@ -146,19 +146,19 @@ static int run_case(void* _pstats)
 
 	gettimeofday(&(pstats->start_time), NULL);
 
-	while(1){
+	do {
 		rc = run(pstats);
 		if(rc != TEST_OK) {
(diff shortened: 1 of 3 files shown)


test/test_multithread_stress.c
```diff
@@ -272,9 +279,39 @@ int main(int argc, char **argv)
 		__atomic_load(&failed_thread, &abort, __ATOMIC_RELAXED);
 	}
 
-	for (int i = 0; i < thread_num; i++) {
-		pthread_join(thread_info[i].tid, NULL);
-	}
+	int try_again=0;
+	gettimeofday(&start_time, NULL);
+	do {
```
mscastanho, line 284: @tuliom  The code will only get here after all threads have returned, so this timeout logic won't help much here. This loop should be combined with the previous one (`while (abort == 0)`) to have the desired effect.

mscastanho (changes requested):

test/test_multithread_stress.c
tuliom: @mscastanho I disagree with you both when you say this logic won't help much and when you say this will be reached only after all threads have returned.
Line 284 is reached in 2 scenarios:

1. When at least 1 thread fails. In this case, all other threads are likely running.
2. When all threads have been **started** `test_iterations` times. In this case, we have between 0 and all threads still running.

In both cases, it does make sense to kill the remaining threads if they do not complete in 5 minutes, but I'm more interested on case 1 because I've seen it triggering issues where threads do not exit after hours of execution.
Case 2 is just expected behavior and, as such, should be completed in less than 5 minutes too.

I do agree with you this is not helping against an uncaught issue in lines 262-272, but I'm afraid it's hard to set such a value when the number of threads, test interval and the number of iterations vary. IMHO, this would be better served with a watchdog monitoring each test independently, i.e. each libnxz test process.
This will require further work and it is not my goal at the moment.

mscastanho (approved):

test/test_multithread_stress.c
mscastanho: Makes sense, thanks for the clarification.

PR #195, opened 2022-08-16
Stop requiring the presence of the device tree

tuliom:
The device tree provides a lot of information besides the amount of NX
GZIP engines available in a system, making it a sensitive directory to
be shared inside a restrictive container.

Add an alternative mechanism to detect the virtualization environment
when the device tree is not available and enumerate 1 NX device if the
NX GZIP device is available and accessible.

### DIFF
--- a/lib/nx_zlib.c
+++ b/lib/nx_zlib.c
@@ -746,6 +746,8 @@ int nx_read_credits(int *total, int *used) {
    Populate NX structures and return number of NX units
 */
 #define DEVICE_TREE "/proc/device-tree"
+#define LPARCFG_FILE "/proc/powerpc/lparcfg"
+#define NX_GZIP_DEV "/dev/crypto/nx-gzip"
 static int nx_enumerate_engines()
 {
 	DIR *d;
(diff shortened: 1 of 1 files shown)


rzinsly (approved):

lib/nx_zlib.c
```diff
@@ -761,11 +763,26 @@ static int nx_enumerate_engines()
...
 	if (d == NULL){
-		prt_err("open device tree dir failed.\n");
+		prt_info("open device tree dir failed.\n");
+
+		/* In a restricted container environment, the device tree
+		   directory may not be available. We can still find out if
+		   this is a PowerVM environment.  */
+		if (access(LPARCFG_FILE, F_OK) == 0) {
+			nx_config.virtualization = POWERVM;
+			return 1;
```
mscastanho, line 773: @tuliom What if the code runs in a PowerVM system which doesn't have `/dev/crypto/nx-gzip` configured? Isn't it better to just set the virtualization type here and leave the number of engines to be decided by the next check?

tuliom: @mscastanho I was trying to avoid this because I didn't want to conflict with the code in `open_device_nodes()` which properly check for the existence, permission and being able to open this device.
Notice the current code for PowerVM returns 1 device without checking if `/dev/crypto/nx-gzip` exists. It seems to return that even on Power9, which is not supported.

If we change this code in order to check for at least existence and permission, should we return 0 if any of them fail?

mscastanho: Ok, if the previous code was already returning 1 I think it's better to leave it as-is for now. We know things are working, let's not mess up this late in the game. :shipit:

mscastanho (approved):

Issue #191, opened 2022-08-10
Build Failure with gcc 11.3.0


amock:
Running
```
./configure
make
```
fails with
```
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I.. -I../inc_nx -DZLIB_PATH=\"libz.so\" -O3 -Wall -Werror -g -mcpu=power9 -MT gzip_vas.lo -MD -MP -MF .deps/gzip_vas.Tpo -c gzip_vas.c  -fPIC -DPIC -o .libs/gzip_vas.o
In file included from gzip_vas.c:68:
nx_zlib.h:132:32: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘OF’
  132 | extern const char *zlibVersion OF((void));
      |                                ^~
```

This is on a big endian system.  I'm able to build this fine under a little endian system.   If I add `#define OF(x) ()` to various places to get the build to succeed then some of the tests fail.

tuliom:
To the best of my knowledge, nobody has ever tested this project on a powerpc64-linux system.
@amock do you have `/dev/crypto/nx-gzip` in this big endian machine?

amock:
@tuliom I do have `/dev/crypto/nx-gzip` and I'm running kernel 5.18.

Do you know where that macro is supposed to come from?  I would at least expect the code to compile successfully on a big endian machine.

amock:
I found that `test_crc32` fails when more than 30 bytes are processed, which lead me to the line https://github.com/libnxz/power-gzip/blob/e1d0b1c7b0df47acdbcc8e3c5fb0ca4d9f28f018/lib/crc32_ppc.c#L70 so `__crc32_vpmsum` is broken.  For some tests I tried (strings of 'a' of length up to 100) changing https://github.com/libnxz/power-gzip/blob/b589fb633e2b13eb1f2429ca2853ca87c2747c5f/lib/crc32_power.c#L506 to 
```c
return v0[0];
```
and the test passed.  That doesn't fix all of the existing test cases and I don't understand the code so I don't know why it works sometimes.

amock:
The test summary:
```
PASS: test_adler32
PASS: test_buf_error
FAIL: test_crc32
../test-driver: line 107: 114513 Aborted                 "$@" > $log_file 2>&1
FAIL: test_inflatesyncpoint
FAIL: test_multithread_stress
PASS: test_pid_reuse
SKIP: test_abi
[...]

PR #196, opened 2022-08-28
Add support for big endian crc32

amock:
This is copied from [1] with minor changes since no support is
needed for reflected crcs, just big-endian and little-endian.

This is work towards fixing #191 

[1] https://github.com/antonblanchard/crc32-vpmsum/

Signed-off-by: Alan Mock <alan@alanmock.com>


lib/crc32_power.c
```diff
@@ -502,6 +551,9 @@ __crc32_vpmsum(unsigned int crc, const void* p, unsigned long len) {
     /* shift result into top 64 bits of */
     v0 = (__vector unsigned long long)vec_sld((__vector unsigned char)v0,
                                               (__vector unsigned char)vzero, 4);
-
-    return v0[1];
+#ifdef __BIG_ENDIAN__
+    return __builtin_unpack_vector_int128 ((vector __int128_t)(v0), 0);
+#else
+    return __builtin_unpack_vector_int128 ((vector __int128_t)(v0), 1);
+#endif
```
tuliom, line 558: This appears to be wrong is causing ppc64le to fail.

tuliom (changes requested):
@amock , the changes look good to me, but I believe they need a few changes:

1. The second patch is unnecessary. Removing trailing whitespace is a plus.
2. We need your `Signed-off-by` in the commit message showing that you agree with https://developercertificate.org.
3. In the commit message, could you elaborate if this implementation is coming from crc32-vpmsum. I noticed some differences and I'm not sure if this is a copy or not.
4. This patch is causing a regression on ppc64le.  Could you check it, please?

amock:
@tuliom thanks for the review.  I think I have made all the requested changes.

lib/crc32_power.c
```diff
@@ -120,6 +130,7 @@ __crc32_vpmsum(unsigned int crc, const void* p, unsigned long len) {
 
         vconst1 = vec_ld(offset, vcrc_short_const);
         vdata0 = vec_ld(0, (__vector unsigned long long*) p);
+        VEC_PERM(vdata0, vdata0, vconst1, vperm_const);
```
RajalakshmiSR, line 133: @amock vperm_const is defined only for big-endian. Do we need it for LE, even though VEC_PERM is an empty macro?

amock: We do not need it since it vanishes with the empty macro on LE.

tuliom (approved):
@amock I'm apologize for taking so long to review it.
LGTM.
@RajalakshmiSR , any other comments?

RajalakshmiSR:
> LGTM.
> @RajalakshmiSR , any other comments?

I merged it, Thanks.

PR #206, opened 2024-01-09
202804 - Uncompress error -5

sachinmonga64:
Testcase fix to release the resources in case the thread fails.


sachinmonga64:
1. There is a script inside library compdecomp_th  which does stress testing in which around 100 threads are created where a fixed data is compression/decompressed parallelly from multiple threads.
2. If in case any thread fails (For example not enough memory left to perform compression), then test case comes in hang state and few threads get stuck alive.
3. So to release the resources these threads have acquired , we need pthread_detach( ) so that the thread becomes separated from the group and it gets terminated after releasing the resources

PR #209, opened 2024-01-24
Fixes for GCC 14 and zlib-ng compat usage

tuliom:
These are trivial fixes.

I verified they compile, but I can't run the tests because I don't have access to systems with NX GZIP anymore. Could you run the tests please?

### DIFF
--- a/lib/nx_crc.c
+++ b/lib/nx_crc.c
@@ -45,8 +45,6 @@
 typedef size_t   z_size_t;
 
 #undef NOBYFOUR
-#define ZEXPORT
-#define ZEXTERN extern
 #define ZSWAP32(q) bswap_32(q)
 
 
--- a/lib/nx_gzlib.c
+++ b/lib/nx_gzlib.c
@@ -142,7 +142,7 @@ gzFile __gzopen(const char* path, int fd, const char *mode, int force_nx)
 
 	state = malloc(sizeof(struct gz_state));
 	if (state == NULL)
-		return NULL;
+		return Z_NULL;
 	memset(state, 0, sizeof(struct gz_state));
 
 	if (path == NULL){
@@ -192,7 +192,7 @@ gzFile __gzopen(const char* path, int fd, const char *mode, int force_nx)
 		if (err == Z_OK) {
 			state->buf = malloc(BUF_LEN);
 			if (state->buf == NULL) {
-				err=Z_NULL;
+				err=Z_MEM_ERROR;
 			}
 		}
 	}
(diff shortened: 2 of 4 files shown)


sachinmonga64:
@tuliom 
We need to run 3 tests ?
1. test
2. serial_test
3. oct

tuliom:
> @tuliom We need to run 3 tests ?
> 
>     1. test
>     2. serial_test
>     3. oct

@sachinmonga64 Ideally, yes. You can run them with `make -j$(nproc) check`, `make -j$(nproc) selftest` and `make -j$(nproc) oct`.
The last one takes longer and is usually skipped.
Considering these changes should not affect the output of the library, I believe it's OK to skip it.

sachinmonga64:
@tuliom  Thanks for the information.

In addition to the changes committed for the id [f244b22] , I think this change may be required at one more location at line:144 in same file

]$ less lib/nx_gzlib.c
```
...
135 gzFile __gzopen(const char* path, int fd, const char *mode, int force_nx)
136 {
137         gz_statep state;
138         gzFile file;
139         int err, strategy = Z_DEFAULT_STRATEGY, level = Z_DEFAULT_COMPRESSION;
140         char* digit;
141         void* stream;
142 
143         state = malloc(sizeof(struct gz_state));
144         if (state == NULL) {
145                 err=Z_MEM_ERROR;
146                 return Z_NULL;
147         }
...
```

Regarding testing, I have done the testing on RHEL8.9 and below is the result for tests directory.

```
...
]$ make check

SKIP: test_window_suspend
SKIP: test_shared_window_suspend
============================================================================
Testsuite summary for libnxz 0.64
============================================================================
# TOTAL: 3
# PASS:  1
# SKIP:  2
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
```

Selftest is good.

For my information , can you please let me know in what situations, these 2 tests(test_window_suspend, test_shared_window_suspend) get SKIP ?

sachinmonga64:
@tuliom  Can you please confirm the result shared in previous comment ?

tuliom:
> For my information , can you please let me know in what situations, these 2 tests(test_window_suspend, test_shared_window_suspend) get SKIP ?

@sachinmonga64 It has to be one of these 3 cases here: https://github.com/libnxz/power-gzip/blob/master/test/serial-tests/test_window_suspend.c#L65-L80

tuliom:
> In addition to the changes committed for the id [[f244b22](https://github.com/libnxz/power-gzip/commit/f244b223c7e4be82c7ec69d76d03d45220b4b754)] , I think this change may be required at one more location at line:144 in same file. I have done testing including this.

Nice catch!
I've just updated the PR including this change.

Thanks!

sachinmonga64:
Hi @tuliom 

Libnxz testing is done from my side. 
Is the testing done on your side with the updated patch that you committed again ?

tuliom:
> Libnxz testing is done from my side. Is the testing done on your side with the updated patch that you committed again ?

@sachinmonga64 Yes.

Issue #203, opened 2023-07-06
EOFException when java nxgzip testing


chenhealth1234:
OS：Centos 8.5
kernel：4.18.0-348.el8.ppc64le
jdk version：jdk-11.0.19-7

when java nxgzip-v0.64 testing, we met EOF issue, the issue is as follow:

LD_PRELOAD=./libnxz.so.0.0.64 java ShortRead
size=14 csize=19 available=14
Exception in thread "main" java.io.EOFException: Unexpected end of ZLIB input stream
        at java.base/ja4va.util.zip.ZipFile$ZipFileInflaterInputStream.fill(ZipFile.java:467)
        at java.base/java.util.zip.InflaterInputStream.read(InflaterInputStream.java:159)
        at java.base/java.io.FilterInputStream.read(FilterInputStream.java:107)
        at ShortRead.main(ShortRead.java:64)

As this issue occur, we did libz and nxgzip-version0.63 testing, found this issue not exist, so we check the data and found the data by nxgzip0.64 is not right, the detail is as follow:

Libz testing:

# cd jdk11u-jdk-11.0.19-7/test/jdk/java/util/zip/ZipFile/
# javac ShortRead.java

# java ShortRead
size=14 csize=16 available=14
uncompressdata:Data disponibl
# xxd abc.zip
00000000: 504b 0304 1400 0808 0800 0e75 dd56 0000  PK.........u.V..
00000010: 0000 0000 0000 0000 0000 0300 0000 6162  ..............ab
00000020: 6373 492c 4954 48c9 2c2e c8cf cb4c ca01  csI,ITH.,....L..
00000030: 0050 4b07 084f ea5b 2f10 0000 000e 0000  .PK..O.[/.......
00000040: 0050 4b01 0214 0014 0008 0808 000e 75dd  .PK...........u.
00000050: 564f ea5b 2f10 0000 000e 0000 0003 0000  VO.[/...........
00000060: 0000 0000 0000 0000 0000 0000 0000 0061  ...............a
00000070: 6263 504b 0506 0000 0000 0100 0100 3100  bcPK..........1.
00000080: 0000 4100 0000 0000                      ..A.....

nxgzip 0.63 testing:
# LD_PRELOAD=./libnxz.so.0.0.63 java ShortRead
size=14 csize=16 available=14
uncompressdata:Data disponibl
# xxd abc.zip
00000000: 504b 0304 1400 0808 0800 0e75 dd56 0000  PK.........u.V..
[...]

PR #215, opened 2026-02-16
nx_inflate.c: Submit job to nx when flush parameters are Z_{FULL,PART…IAL}_FLUSH [#203]

avithemad:
After commit 94573fe, which allowed (de)compression of streams even if avail_in < threshold, a latent bug in inflate logic was exposed, which was found during integration test with open-jdk.

This patch fixes this issue in nx_inflate_ by adding checks to see if the flush is also not Z_{FULL,PARTIAL}_FLUSH along with Z_FINISH and Z_SYNC_FLUSH, when pushing the data to fifo.

### DIFF
--- a/lib/nx_inflate.c
+++ b/lib/nx_inflate.c
@@ -1197,7 +1197,10 @@ copy_fifo_out_to_next_out:
 	if (s->avail_in > 0
 	    && (s->avail_in + s->used_in < nx_config.cache_threshold)
 	    && s->avail_out > 0
-	    && flush != Z_FINISH && flush != Z_SYNC_FLUSH) {
+	    && flush != Z_FINISH
+	    && flush != Z_FULL_FLUSH
+	    && flush != Z_SYNC_FLUSH
+	    && flush != Z_PARTIAL_FLUSH) {
 		/* We haven't accumulated enough data. Cache any input data
 		   provided and wait for the application to send more in order
 		   to reduce the amount of requests sent to the accelerator. */


avithemad:
@sachinmonga64 @tuliom 
Requesting review.

tuliom:
@avithemad I'm sorry, but I'm afraid I'm unable to properly review PRs for libnxz.
I do not have access to a system with NX GZIP anymore.

I see that @sachinmonga64 has full access to the repo in order to review and merge PRs.

avithemad:
@tuliom Sure. I wanted to ask a doubt regarding the fix in 
[94573fedddd46a24687f0d6489f434cb1333b3f7](https://github.com/libnxz/power-gzip/commit/94573fedddd46a24687f0d6489f434cb1333b3f7)

Here as mentioned in the commit message, that the first call to inflate/deflate may not have enough data even when handling large data. Under what scenarios can this happen?
Also there are different flush values like Z_FULL_FLUSH and Z_PARTIAL_FLUSH. For these values as well even if the data is small it goes through nx route, when data is small. 
The change in this PR is required for correctness. But should we also add those in here:
```
	if((flush == Z_FINISH || flush == Z_FULL_FLUSH || flush == Z_PARTIAL_FLUSH) && strm->avail_in <= DECOMPRESS_THRESHOLD)
```

Your feedback would be greatly appreciated for https://github.com/libnxz/power-gzip/pull/219, since we are missing a bit of context on the commit 94573fedddd46a24687f0d6489f434cb1333b3f7. 

Im abandoning this PR, since we want to merge this in develop branch.

PR #217, opened 2026-02-16
nx_deflate.c: Reset history_len to 0 in deflateResetKeep and deflateInit

avithemad:
The jdk test DeInflate.java exposed this issue where the cur_in=0 and history_len=4096, which caused an assertion failure in nx_setup_ddl_in.

This patch makes sure the state of the stream is consistent when deflateResetKeep of deflateInit is done i.e., initialize/reset the value of history_len to 0 when cur_in is set to 0.


avithemad:
@sachinmonga64 @tuliom 
Requesting review.

avithemad:
Abandoning this PR, since it is requested on develop and not on master.

PR #219, opened 2026-03-09
nx_inflate.c: Submit job to nx when flush parameters are Z_{FULL,PARTIAL}_FLUSH [#203]

avithemad:
After commit 94573fe, which allowed (de)compression of streams even if avail_in < threshold, a latent bug in inflate logic was exposed, which was found during integration test with open-jdk.
This patch fixes this issue in nx_inflate_ by adding checks to see if the flush is also not Z_{FULL,PARTIAL}_FLUSH along with Z_FINISH and Z_SYNC_FLUSH, when pushing the data to fifo.

The util.zip library in open-jdk sends the flush value as `Z_PARTIAL_FLUSH` with the initial call to inflate/deflate. Due to this, the `use_nx_inflate`  function which helps decide whether to use software or nx for the further sequence of inflate/deflate until `Z_FINISH` decides to use nx even for small data.

For example, a simple test case: https://github.com/openjdk/jdk/blob/master/test/jdk/java/util/zip/ZipFile/ShortRead.java
Even for such small data, the open-jdk sends the following sequence for compression
```
nx_deflate:1671 avail_in 15 total_in 0 used_in 0 cur_in 0 avail_out 512 total_out 0 used_out 0 cur_out 0 len_in 0 len_out 4194304 flush 0, actual flush value 1 (Z_PARTIAL_FLUSH)
```

The use_nx_inflate decides to use nx for this case, and since the cache_threshold is not met, it never flushes out of the fifo, and thus triggering the EOFException.

### DIFF
--- a/lib/nx_inflate.c
+++ b/lib/nx_inflate.c
@@ -1188,16 +1188,17 @@ copy_fifo_out_to_next_out:
 	   Avoid executing this code when avail_in is 0.  That means either the
 	   end of the stream or the end of the current request.  There is
 	   nothing to copy anyway.
-	   Likewise when flush is either Z_FINISH or Z_SYNC_FLUSH.  In these
-	   cases, inflate is expected to provide an output and copying data to
-	   fifo_in would just add unnecessary delays.
+	   Cache the input in fifo_in only when flush value is Z_NO_FLUSH.
+	   All other flush modes i.e., Z_SYNC_FLUSH, Z_PARTIAL_FLUSH, Z_FINISH
+	   and Z_FULL_FLUSH require immediate output and therefore must be
+	   submitted to NX and not fifo_in.
 	   The following code is not just an optimization, it is also required
 	   by NX because it may refuse to start processing a stream if the input
 	   is not large enough.  */
 	if (s->avail_in > 0
 	    && (s->avail_in + s->used_in < nx_config.cache_threshold)
 	    && s->avail_out > 0
-	    && flush != Z_FINISH && flush != Z_SYNC_FLUSH) {
+	    && flush == Z_NO_FLUSH) {
 		/* We haven't accumulated enough data. Cache any input data
 		   provided and wait for the application to send more in order
 		   to reduce the amount of requests sent to the accelerator. */


avithemad:
@sachinmonga64 : Kindly review this. It fixes #203 and enables jdk integration which is currently broken due to https://github.com/libnxz/power-gzip/commit/94573fedddd46a24687f0d6489f434cb1333b3f7. 

@tuliom : Your feedback for this PR would be greatly appreciated, i.e., whether this fix is in the right direction, or a better way to handle this?
The main problem is that for Z_PARTIAL_FLUSH, use_nx_inflate returns true. But nx_inflate_ function does not flush the fifo unless it sees Z_FINISH or Z_SYNC_FLUSH

tuliom (changes requested):
This solution seems to be in the right direction, but I suggest a couple of changes:

1. This should be squashed into a single commit;
2. It would be great to describe the openjdk behavior in the commit message because that helps reviewers;
3. There are 7 possible values for `flush`. 4 of them are now listed in this code. Should we modify the code in order to list the other 3 instead? I also wonder if it should just be `&& flush == Z_NO_FLUSH`.

sachinmonga64:
@tuliom - Thanks for your valuable input and I agree with all of your points with just an added comment for point3.

Before adding code for point 3, @avithemad needs to check 
a) if Z_NO_FLUSH ,Z_BLOCK and Z_TREES are really needed at the same location where other 4 FLUSH are listed. (I too will check on this )
b) What is the existing control flow of these FLUSH (Z_NO_FLUSH ,Z_BLOCK and Z_TREES) ?

Just another question is what Flush value is coming in openjdk case ?

avithemad:
Thanks for looking into this.
> This solution seems to be in the right direction, but I suggest a couple of changes:
> 
>     1. This should be squashed into a single commit;
> 
>     2. It would be great to describe the openjdk behavior in the commit message because that helps reviewers;
> 
Sure, I have updated the 1st comment with a simple jdk behaviour.  One concern on this is that even for small data, the library decides to use nx, and I think this was not the original intention. But it is hard to detect these scenarios by just using flush values as done in https://github.com/libnxz/power-gzip/commit/94573fedddd46a24687f0d6489f434cb1333b3f7

>     3. There are 7 possible values for `flush`. 4 of them are now listed in this code. Should we modify the code in order to list the other 3 instead? I also wonder if it should just be `&& flush == Z_NO_FLUSH`.

Yes this makes sense, as the other 2 values Z_TREES and Z_BLOCK are currently not implemented and nx_inflate rejects these values:
https://github.com/libnxz/power-gzip/blob/2f1ae5450cac8cc5f93941fd099fbe319e55a319/lib/nx_inflate.c#L304
will update the code for this.

avithemad:
@tuliom , @sachinmonga64 , updated the branch to just check if flush == Z_NO_FLUSH. Also updated the comment that described this if block.

avithemad:
Update: the actual flush value being passed was Z_PARTIAL_FLUSH, but I was confused as the logs only show flush value of the nx_streamp, which is not updated for the stream.
Should we add the line:
```
	/* update the flush status here */
	s->flush = flush;
```
right after here:
https://github.com/libnxz/power-gzip/blob/2f1ae5450cac8cc5f93941fd099fbe319e55a319/lib/nx_inflate.c#L323

Similar to how its done in nx_deflate.c

tuliom:
> Sure, I have updated the 1st comment with a simple jdk behaviour. One concern on this is that even for small data, the library decides to use nx, and I think this was not the original intention.

@avithemad I see you updated the description (aka. first comment) in this PR.
I'm afraid this will not end up in the git history though. I suspect that git history will end up exactly as it is now in your branch, i.e. with 3 commits.

That's why I suggested to squash the commits.

avithemad:
> @avithemad I see you updated the description (aka. first comment) in this PR. I'm afraid this will not end up in the git history though. I suspect that git history will end up exactly as it is now in your branch, i.e. with 3 commits.
 
Got it. Apologies, as I was under the impression that squash can happen during the merge of the pull request, as this was the previous workflow I was aware of.

> That's why I suggested to squash the commits.

I have squashed the commits, updated the commit message with the jdk scenario described in there as well.
Thanks for reviewing.

tuliom (approved):
LGTM

sachinmonga64:
>Got it. Apologies, as I was under the impression that squash can happen during the merge of the pull request, as this was the previous workflow I was aware of.

squash can be done either ways but it is better if we do it from contributor side. Thanks for doing it.

@tuliom - Thanks.

Will take a look

lib/nx_inflate.c
```diff
@@ -1188,16 +1188,16 @@ static int nx_inflate_(nx_streamp s, int flush)
 	   Avoid executing this code when avail_in is 0.  That means either the
 	   end of the stream or the end of the current request.  There is
 	   nothing to copy anyway.
-	   Likewise when flush is either Z_FINISH or Z_SYNC_FLUSH.  In these
-	   cases, inflate is expected to provide an output and copying data to
-	   fifo_in would just add unnecessary delays.
+	   Likewise when flush is either Z_FINISH, Z_SYNC_FLUSH Z_PARTIAL_FLUSH
```
sachinmonga64: a) Missing Comma
b) Rather can we update the comment to match the code logic. The current comment lists specific 
flush modes, but the code now simply checks `flush == Z_NO_FLUSH`
Simplifying comment something like

/* Cache data in FIFO only when Z_NO_FLUSH is specified. All other
   flush modes (Z_SYNC_FLUSH, Z_PARTIAL_FLUSH, Z_FULL_FLUSH, Z_FINISH)
   require immediate output and should submit to NX rather than buffer. */

sachinmonga64 (commented):
@avithemad 

Apart from comment change, can we add a test(regression) case to libnxz test-suite for Z_PARTIAL_FLUSH flush  ?

lib/nx_inflate.c
avithemad: Thanks for the comments. I have updated the comments as requested.

Will add test for this in a separate PR.

sachinmonga64 (approved):
LGTM

Ready to Merge.

sachinmonga64:
But as discussed, please add a corresponding testcase for verifying new flush type in a separate patch.

PR #222, opened 2026-04-22
Enhancements/Corrections in existing logging

avithemad:
There were a few inconsistencies in logging in existing code. This PR fixes them.

### DIFF
--- a/lib/nx_deflate.c
+++ b/lib/nx_deflate.c
@@ -624,8 +624,13 @@ int nx_deflateInit2_(z_streamp strm, int level, int method, int windowBits,
 	else wrap = HEADER_ZLIB;
 
 	prt_info(" windowBits %d wrap %d \n", windowBits, wrap);
-	if (method != Z_DEFLATED || (strategy != Z_FIXED && strategy != Z_DEFAULT_STRATEGY)) {
-		prt_err("unsupported zlib method or strategy\n");
+	if (method != Z_DEFLATED) {
+		prt_err("unsupported zlib method. Supported zlib methods: Z_DEFLATED.\n");
+		return Z_STREAM_ERROR;
+	}
+	if (strategy != Z_FIXED && strategy != Z_DEFAULT_STRATEGY) {
+		prt_err("unsupported zlib strategy."
+			" Supported zlib strategies: Z_FIXED, Z_DEFAULT_STRATEGY.\n");
 		return Z_STREAM_ERROR;
 	}
 
@@ -1329,12 +1334,12 @@ restart:
 				/* When page faults are too many oom_killer
 				 * should kill this process. */
 				rc = LIBNX_ERR_PAGEFLT;
-				prt_err("Cannot make progress; ");
-				prt_err("too many page faults!\n");
+				prt_err("Cannot make progress; too many page"
+					" faults cc= %d\n", cc);
 				goto err_exit;
 			}
 			else {
-				prt_warn("ERR_NX_AT_FAULT: Retry again\n");
+				prt_warn("ERR_NX_AT_FAULT: Retry\n");
 				goto restart;
 			}
 		}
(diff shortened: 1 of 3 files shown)


sachinmonga64 (commented):
Can we change the log message to 

The flush value in the nx_stream was not updated in nx_inflate,  which could lead to functional incorrectness or inconsistent behavior. This commit adds the line to update it, similar to how its done in nx_deflate.c

Otherwise LGTM

lib/nx_deflate.c
```diff
@@ -1691,7 +1696,8 @@ int nx_deflate(z_streamp strm, int flush)
 
 	/* User must not provide more input after the first FINISH: */
 	if (s->status == NX_BFINAL_ST && s->avail_in != 0) {
-		prt_info("s->status is NX_BFINAL_ST but s->avail_out is not 0\n");
+		prt_info("s->status = NX_BFINAL_ST, s->avail_in = %d"
+			 " Expected s->avail_in to be 0.\n", s->avail_in);
```
sachinmonga64: 1. Can we write something which gives more clarity like "Non-zero s->avail_in value" 

prt_info("s->status is NX_BFINAL_ST but s->avail_out is not 0\n");
		prt_info("s->status = NX_BFINAL_ST, s->avail_in = %d"
			 " Non Zero s->avail_in value.\n", s->avail_in);

2.  Shouldn't this be prt_err instead of prt_info ? Even if earlier it was prt_info , we should change it. Think about it.

lib/nx_inflate.c
```diff
@@ -1366,7 +1366,7 @@ static int nx_inflate_(nx_streamp s, int flush)
 					" faults cc= %d\n", cc);
 			}
 			else {
-				prt_warn("ERR_NX_AT_FAULT: more retry\n");
+				prt_warn("ERR_NX_AT_FAULT: Retry again\n");
```
sachinmonga64: Can we write only "Retry" ?

prt_warn("ERR_NX_AT_FAULT: Retry\n");

lib/nx_zlib.c
```diff
@@ -731,9 +731,12 @@ static int nx_query_job_limits()
 }
 
 int nx_read_credits(int *total, int *used) {
-	if (nx_read_sysfs_entry(SYSFS_VAS_CAPS "nr_total_credits", total) ||
-	    nx_read_sysfs_entry(SYSFS_VAS_CAPS "nr_used_credits", used)) {
-		prt_info("Failed to read number of credits from sysfs.\n");
+	if (nx_read_sysfs_entry(SYSFS_VAS_CAPS "nr_total_credits", total)) {
+		prt_info("Failed to read number of total credits from sysfs.\n");
```
sachinmonga64: prt_err instead of prt_info ?

lib/nx_zlib.c
```diff
@@ -731,9 +731,12 @@ static int nx_query_job_limits()
...
+		return -1;
+	}
+	if (nx_read_sysfs_entry(SYSFS_VAS_CAPS "nr_used_credits", used)) {
+		prt_info("Failed to read number of used credits from sysfs.\n");
```
sachinmonga64: prt_err instead of prt_info ?

sachinmonga64 (changes requested):

lib/nx_deflate.c
avithemad: Sure, agreed on changing the message. 

But for the second part, looked around the code, and it looks like for Z_BUF_ERROR, prt_info is being used.
One reasoning I can think of is in the zlib manual mentions this
```
Note that Z_BUF_ERROR is not fatal, and deflate() can be called again with more input and more output space to continue compressing. 
```
So, I think we should use prt_err only for fatal errors like Z_STREAM_ERROR?

lib/nx_inflate.c
avithemad: Done

lib/nx_zlib.c
avithemad: Done

lib/nx_zlib.c
avithemad: Done

avithemad:
> Can we change the log message to
> 
> The flush value in the nx_stream was not updated in nx_inflate, which could lead to functional incorrectness or inconsistent behavior. This commit adds the line to update it, similar to how its done in nx_deflate.c
> 
> Otherwise LGTM

Thanks for the review. I have update this specific commit message.

lib/nx_deflate.c
sachinmonga64: Thanks Avinash.

You're correct. I mistook it Z_BUF_ERROR  for Z_STREAM_ERROR
Let's keep it prt_info only.

sachinmonga64 (approved):

sachinmonga64 (approved):
Looks Good

PR #223, opened 2026-04-23
Add public endpoint tracing capability for better debugging

avithemad:
It is hard to reproduce the test cases just by observing the logs right
    now, specially in cases where other languages like Java uses this
    library.
    This commit tries to make it easier to debug in such scenarios by adding
    logs at each public api entry point. We can further use the log to
    generate a test case in C, if required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,15 @@ Use `export NX_GZIP_TRACE=1` to enable logic trace.
 
 Use `export NX_GZIP_TRACE=8` to enable statistics trace.
 
+Sometimes, a sequence of public api call sequence may be needed to reproduce
+some issues such as when the library is used with other languages/sdk like
+JDK. In those situations, build the library with public api trace support
+```
+./configure CFLAGS="-DNX_API_TRACE"
+```
+Then use `export NX_GZIP_VERBOSE=2` to observe the sequence of public api calls
+in the log file.
+
 ## Supported Functions List
 
 All the zlib supported functions are listed and described at libnxz.h.
--- a/inc_nx/nx_dbg.h
+++ b/inc_nx/nx_dbg.h
@@ -112,6 +112,15 @@ extern pthread_mutex_t mutex_log;
 	prt("### "fmt, ## __VA_ARGS__);					\
 }} while (0)
 
+/* Trace public API entry points - compile-time conditional for zero overhead */
+#ifdef NX_API_TRACE
+#define prt_api_entry(func, fmt, ...) \
+	prt_info(">>> API ENTRY: %s(" fmt ")\n", func, ## __VA_ARGS__)
+#else
+#define prt_api_entry(func, fmt, ...)
+#endif
+
+
 /**
  * str_to_num - Convert string into number and copy with endings like
  *              KiB for kilobyte
--- a/lib/nx_deflate.c
+++ b/lib/nx_deflate.c
@@ -2275,6 +2275,11 @@ int deflateInit2_(z_streamp strm, int level, int method, int windowBits,
 	void *sw_state = NULL;
 	nx_streamp s;
 
+	prt_api_entry("deflateInit2_", "strm=%p, level=%d, method=%d, windowBits=%d,"
+			"memLevel=%d, strategy=%d, version=%s, stream_size=%d",
+			strm, level, method, windowBits, memLevel, strategy, version, stream_size);
+	print_zstream_info(strm, __LINE__);
+
 	/* statistic */
 	zlib_stats_inc(&zlib_stats.deflateInit);
 
@@ -2357,6 +2362,9 @@ int deflateInit2_(z_streamp strm, int level, int method, int windowBits,
 int deflateReset(z_streamp strm)
 {
 	int rc;
+
+	prt_api_entry("deflateReset", "strm=%p", strm);
+	print_zstream_info(strm, __LINE__);
 
 	if (nx_config.mode.deflate == GZIP_AUTO) {
 		struct stream_map_entry *sme;
@@ -2454,6 +2462,8 @@ int deflateReset(z_streamp strm)
 */
 int deflateResetKeep(z_streamp strm)
 {
+	prt_api_entry("deflateResetKeep", "strm=%p", strm);
+	print_zstream_info(strm, __LINE__);
 	if (has_nx_state(strm))
 		return nx_deflateResetKeep(strm);
 	else
@@ -2463,6 +2473,8 @@ int deflateResetKeep(z_streamp strm)
 int deflateEnd(z_streamp strm)
 {
 	int rc;
+	prt_api_entry("deflateEnd", "strm=%p", strm);
+	print_zstream_info(strm, __LINE__);
 
 	/* statistic */
 	zlib_stats_inc(&zlib_stats.deflateEnd);
@@ -2508,6 +2520,9 @@ int deflate(z_streamp strm, int flush)
 	unsigned int avail_in_slot, avail_out_slot;
 	uint64_t t1=0, t2, t_diff;
 	unsigned int avail_in=0, avail_out=0;
+
+	prt_api_entry("deflate", "strm=%p, flush=%d", strm, flush);
+	print_zstream_info(strm, __LINE__);
 
 	/* statistic */
 	if (nx_gzip_gather_statistics()) {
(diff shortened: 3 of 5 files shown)


inc_nx/nx_dbg.h
```diff
@@ -112,6 +112,15 @@ extern pthread_mutex_t mutex_log;
 	prt("### "fmt, ## __VA_ARGS__);					\
 }} while (0)
 
+/* Trace public API entry points - compile-time conditional for zero overhead */
+#ifdef NX_API_TRACE
+#define prt_api_entry(func, fmt, ...) \
```
sachinmonga64, line 117: Why we need func as an argument when `__func__` is already present ? Adding function naming separately will add unnecessary overhead.

inc_nx/nx_dbg.h
```diff
@@ -112,6 +112,15 @@ extern pthread_mutex_t mutex_log;
...
+	prt_info(">>> API ENTRY: %s(" fmt ")\n", func, ## __VA_ARGS__)
```
sachinmonga64, line 118: fmt within brackets look inconsistent with existing surrounding code

inc_nx/nx_dbg.h
```diff
@@ -112,6 +112,15 @@ extern pthread_mutex_t mutex_log;
...
+#else
+#define prt_api_entry(func, fmt, ...)
+#endif
+
+
```
sachinmonga64, line 123: Trivial but remove the extra new line.

inc_nx/nx_dbg.h
sachinmonga64: The below version looks clean.

#ifdef NX_API_TRACE
#define prt_api_entry(fmt, ...) \
        prt_info(">>> API ENTRY: %s: " fmt, `__func__`, ## `__VA_ARGS__`)
#else
#define prt_api_entry(fmt, ...)
#endif

lib/nx_deflate.c
```diff
@@ -2275,6 +2275,11 @@ int deflateInit2_(z_streamp strm, int level, int method, int windowBits,
 	void *sw_state = NULL;
 	nx_streamp s;
 
+	prt_api_entry("deflateInit2_", "strm=%p, level=%d, method=%d, windowBits=%d,"
```
sachinmonga64, line 2278: Change the argument as per comment #1 if we are dropping func argument. Same for all following callings.

lib/nx_zlib.h
```diff
@@ -524,6 +524,18 @@ len_in %ld len_out %ld flush %d\n", __FUNCTION__, line, \
 (long)(s)->len_in, (long)(s)->len_out, (s)->flush);	\
 } while (0)
 
+/* Print z_stream state information for API tracing - compile-time conditional for zero overhead */
+#ifdef NX_API_TRACE
+#define print_zstream_info(strm, line) \
+do { if (strm) prt_info(\
+"== %s:%d z_stream: avail_in %u total_in %lu avail_out %u total_out %lu\n", \
+__FUNCTION__, line, \
```
sachinmonga64, line 532: 1. I think it will be better to use `__func__` instead of `__function__` ??

2. Why we need to pass `line` when `__LINE__` is already available inside the function ?

sachinmonga64: #define print_zstream_info(strm) \
    do { if (strm) prt_info( \
        "== %s:%d z_stream: avail_in %u total_in %lu avail_out %u total_out %lu\n", \
        `__func__`, `__LINE__`, \
        (strm)->avail_in, (strm)->total_in, (strm)->avail_out, (strm)->total_out); \
    } while (0)

README.md
```diff
@@ -48,6 +48,15 @@ Use `export NX_GZIP_TRACE=1` to enable logic trace.
 
 Use `export NX_GZIP_TRACE=8` to enable statistics trace.
 
+Sometimes, a sequence of public api call sequence may be needed to reproduce
```
sachinmonga64, line 51: More or less your message is delivered correctly in README.md file.

However , I want to add few add-ons and also a bit of change in framing. So I think the below message delivers it well. Add it as it is

>8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8

Sometimes, reproducing issues requires observing the sequence of public API
calls — especially when libnxz is used via other languages or SDKs such as JDK.

To enable API entry tracing, build the library with:

    ./configure CFLAGS="-DNX_API_TRACE"

Then set `export NX_GZIP_VERBOSE=2` at runtime to observe the public API
call sequence in the log output.

Note: NX_API_TRACE is a compile-time flag. Without it, API tracing adds
zero overhead to the library.

>8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8 >8

lib/nx_deflate.c
```diff
@@ -2454,6 +2462,8 @@ int deflateReset(z_streamp strm)
 */
 int deflateResetKeep(z_streamp strm)
 {
+	prt_api_entry("deflateResetKeep", "strm=%p", strm);
+	print_zstream_info(strm, __LINE__);
```
sachinmonga64, line 2466: Do we need a blank line after print_zstream_info to separate it from function logic ? I think this is the consistency followed everywhere

lib/nx_deflate.c
```diff
@@ -2559,6 +2574,8 @@ unsigned long deflateBound(z_streamp strm, unsigned long sourceLen)
 {
 	unsigned long rc;
 
+	prt_api_entry("deflateBound", "strm=%p, sourceLen=%lu", strm, sourceLen);
+	print_zstream_info(strm, __LINE__);
```
sachinmonga64, line 2578: same. Can we add a blank line ?

lib/nx_deflate.c
```diff
@@ -2575,6 +2592,8 @@ unsigned long deflateBound(z_streamp strm, unsigned long sourceLen)
 
 int deflateParams(z_streamp strm, int level, int strategy)
 {
+	prt_api_entry("deflateParams", "strm=%p, level=%d, strategy=%d", strm, level, strategy);
+	print_zstream_info(strm, __LINE__);
```
sachinmonga64, line 2596: Blank line

lib/nx_deflate.c
```diff
@@ -2594,6 +2613,8 @@ int deflateParams(z_streamp strm, int level, int strategy)
 
 int deflateSetHeader(z_streamp strm, gz_headerp head)
 {
+	prt_api_entry("deflateSetHeader", "strm=%p", strm);
+	print_zstream_info(strm, __LINE__);
```
sachinmonga64, line 2617: Blank line

lib/nx_inflate.c
```diff
@@ -1990,6 +1990,8 @@ int inflateInit2_(z_streamp strm, int windowBits, const char *version, int strea
 	void *sw_state = NULL;
 	nx_streamp s;
 
+	prt_api_entry("inflateInit2_", "strm=%p, version=%s, stream_size=%d", strm, version, stream_size);
+	print_zstream_info(strm, __LINE__);
```
sachinmonga64, line 1994: This one seems fine as we are collecting stats in the same block

lib/nx_inflate.c
```diff
@@ -2151,6 +2156,8 @@ int inflateReset(z_streamp strm)
 */
 int inflateResetKeep(z_streamp strm)
 {
+	prt_api_entry("inflateResetKeep", "strm=%p", strm);
+	print_zstream_info(strm, __LINE__);
```
sachinmonga64, line 2160: blank line

lib/nx_inflate.c
```diff
@@ -2159,6 +2166,8 @@ int inflateResetKeep(z_streamp strm)
 
 int inflateReset2(z_streamp strm, int windowBits)
 {
+	prt_api_entry("inflateReset2", "strm=%p, windowBits=%d", strm, windowBits);
+	print_zstream_info(strm, __LINE__);
```
sachinmonga64, line 2170: blank line

sachinmonga64 (changes requested):
@avithemad - Please go through the comments

inc_nx/nx_dbg.h
sachinmonga64: Changed

inc_nx/nx_dbg.h
sachinmonga64: Changed

sachinmonga64:
Review fixes have been incorporated and pushed directly to develop (commit a789e46). Closing this PR. 
Thanks for the contribution!

